PR #3, opened 2020-06-24
Setup API endpoints

stevenwjy:
## Summary

The following are the changes made in this PR:
- Add Open API definitions for the PISP demo server to interact with the mobile app and Mojaloop
- Split the communication for mojaloop and app into virtual hosts
- Add bare-bones handlers 
- Fix tests and various checks to pass circle CI

Note:
- Changed the `audit:resolve` and `audit:check` scripts to only evaluate production dependencies. Otherwise, we will need a `audit-resolve.json` file of ~42k lines to ignore the checks related to the dev dependencies (`lodash`, `commitlint`, etc.). 
- Currently the minimum coverage for unit tests is removed (changed from `80` to `10`) since the PR is already looking too huge. I will add it using other PRs as the handlers are being implemented.

## APIs for [pisp-demo-app-flutter](https://github.com/mojaloop/pisp-demo-app-flutter)

How to check:
- All of the path definitions are written in `src/interface/app.yaml`
- To view this, execute `npm run build:openapi:app`. The bundled Open API file will be written on `dist/openapi/app.yaml`
- Copy and paste the bundled file to a swagger editor

TODO (in other PR):
- Serve the swagger documentation so that people don't need to copy-paste the definition to a swagger editor.
- Implement handlers logic
- Add API definitions for polling by the mobile app

## APIs for communication with Mojaloop

How to check:
- All of the path definitions are written in `src/interface/mojaloop.yaml`
- To view this, execute `npm run build:openapi:mojaloop`. The bundled Open API file will be written on `dist/openapi/mojaloop.yaml`
- Copy and paste the bundled file to a swagger editor

TODO (in other PR):
- Serve the swagger documentation so that people don't need to copy-paste the definition to a swagger editor.
- Implement handlers logic
- Confirm the models with what will be used by the third-party adapter.


src/interface/app.yaml
```diff
@@ -0,0 +1,302 @@
...
+  /participants:
+    get:
+      tags:
+        - participants
+      summary: 'Lookup the registered DFSPs in Mojaloop'
+      description: The HTTP request GET /participants is used to find the list of registered DFSPs in Mojaloop
+      operationId: getAppParticipants
+      responses:
+        202:
+          $ref: '../../node_modules/@mojaloop/api-snippets/v1.0/openapi3/responses/index.yaml#/202'
```
looilee, line 25: I'd avoid using relative paths with '../..'. Can you at least put this '../../node_modules' into an environment variable like $NODE_MODULES

src/interface/app.yaml
```diff
@@ -0,0 +1,302 @@
+openapi: 3.0.1
+info:
+  title: PISP Demo Server Open API
+  description: Based on API Definition.docx Version 1.1 that was confirmed on 2020-05-19.
```
looilee, line 4: where is this API Definition.docx file?

stevenwjy:
> Please do not add any more commits to this PR, it is already too big. Please make your PRs more granular in the future.

Okay, got it, actually this PR was intended to only contain changes related to setting up the Open API endpoints, models, and backend. However, since the CircleCI was suddenly activated, I ended up putting all the changes needed to pass the various checks in this PR.

> It would be good to have a readme.md file describing how you organize the source code.

Okay, sure! I will add it using another PR.

src/interface/app.yaml
stevenwjy: Actually, it is also something that I intended to do... However, I don't think it is currently possible to use internal variable in the Open API file.

src/interface/app.yaml
stevenwjy: Oh, I actually just copied the description from the Mojaloop's Open API template, so I believe it refers to Mojaloop API Definition. Shall I just change this into a different description?

looilee: I'd imagine that this is very confusing for someone without context of the template. In fact I couldn't even find that file. Maybe you can remove this.

stevenwjy: okay, updated

PR #4, opened 2020-07-06
Add docs for transfer flow

stevenwjy:
This PR contains sequence diagrams for the transfer flow that capture the interactions between user, PISP App, Firebase, PISP Server, and Mojaloop. 

For ease of review process, the links in the [transfer.md](https://github.com/mojaloop/pisp-demo-server/blob/transfer-sequence-diagram/docs/design/transfer.md) file still refer to the `transfer-sequence-diagram` branch. If the current design looks good, I will update it to point to the master branch before merging.


docs/README.md
```diff
@@ -2,6 +2,11 @@
 
 Documentation for the PISP demo server
 
+## Design
+
+There are two main scenarios that are handled by the PISP demo server:
+- [Transfer](./design/transfer.md)
+
```
looilee, line 9: ?

docs/assets/diagrams/transfer/02-create-transaction.puml
```diff
@@ -0,0 +1,132 @@
...
+end box
+
+Alice -> App: I want to make this transaction
+
+rnote right of App #Light 
+  ""createDoc(""
+  ""  "transactions/users/1234/<uuid>",""
+  ""  {""
+  ""    payee: {""
+  ""      phoneNumber: "+1-222-222-2222"""
```
looilee, line 28: Do you want to include the type i.e. MSISDN?

docs/assets/diagrams/transfer/02-create-transaction.puml
```diff
@@ -0,0 +1,132 @@
...
+App -> Firebase: Create a new document in the database
+return ""200 OK""
+
+rnote right of Firebase #Light
+  ""event(""
+  ""  "document.create",""
+  ""  "transactions/users/1234/<uuid>",""
+  ""  {""
+  ""    payee: {""
+  ""      phoneNumber: "+1-222-222-2222"""
```
looilee, line 44: ditto

docs/assets/diagrams/transfer/02-create-transaction.puml
```diff
@@ -0,0 +1,132 @@
...
+
+Adapter -> Adapter: Lookup data
+
+rnote left of Adapter #Light
+  ""**PUT /parties/{Type}/{ID}**""
+  ""{ ""
+  ""  "party": {""
+  ""    "partyIdInfo": {""
+  ""      "partyIdType": enum,""
+  ""      "partyIdentifier": string,""
```
looilee, line 65: You should fill this in with the example data, i.e. "+1-222-222-2222"

docs/assets/diagrams/transfer/02-create-transaction.puml
```diff
@@ -0,0 +1,132 @@
...
+  ""      "partySubIdOrType": string,""
+  ""      "fspId": number""
+  ""    },""
+  ""    "merchantClassificationCode": string,""
+  ""    "name": string,""
+  ""    "personalInfo": {""
+  ""      "complexName": {""
+  ""        "firstName": string,""
```
looilee, line 73: I think the app displays this? Put in some sample data, I believe it is "Alice"?

docs/assets/diagrams/transfer/02-create-transaction.puml
```diff
@@ -0,0 +1,132 @@
...
+  ""        "middleName": string,""
+  ""        "lastName": string""
```
looilee, line 75: ditto

docs/assets/diagrams/transfer/02-create-transaction.puml
```diff
@@ -0,0 +1,132 @@
...
+deactivate Adapter
+
+rnote left of Server #Light
+  ""updateDoc(""
+  ""  "transactions/users/1234/<uuid>",""
+  ""  {""
+  ""    payee: {""
+  ""      name: "Alice Alpaca",""
+  ""      fspId: "fspa",""
+  ""      phoneNumber: "+1-222-222-2222"""
```
looilee, line 102: do you need type?

docs/assets/diagrams/transfer/03a-confirm-payee.puml
```diff
@@ -0,0 +1,127 @@
...
+  participant "Third-party\nAPI Adapter" as Adapter
+end box
+
+Alice -> App: The payee info is correct
+
+rnote right of App #Light
+  ""updateDoc(""
+  ""  "transactions/users/1234/<uuid>",""
+  ""  {""
+  ""    sourceAccountId: "bob.fspa",""
```
looilee, line 27: Does sourceAccountId come from linking? Where is it persisted in the app?

docs/assets/diagrams/transfer/03a-confirm-payee.puml
```diff
@@ -0,0 +1,127 @@
...
+Firebase -> Server ++: Notify update on transaction status
+Firebase <-- Server: ""200 OK""
+
+rnote right of Server #Light 
+  ""**POST /thirdpartyRequests/transactions**""
+  ""FSPIOP-Source: ...""
+  ""FSPIOP-Destination: ...""
+  ""{""
+  ""  "transactionRequestId": string,""
+  ""  "sourceAccountId": string,""
```
looilee, line 66: Is every value here empty?

docs/README.md
stevenwjy: The other point should be:
```
- [Linking](./design/linking.md)
```

However, it is currently written only on the other draft PR, do you think it's better to put that one here as well? Or perhaps I can change the phrasing to:

```
Here are the scenarios that are handled by the PISP Demo server:
```

stevenwjy: Updated the phrasing to the second option

docs/assets/diagrams/transfer/02-create-transaction.puml
stevenwjy: updated

docs/assets/diagrams/transfer/02-create-transaction.puml
stevenwjy: updated

docs/assets/diagrams/transfer/03a-confirm-payee.puml
stevenwjy: Yes, I think for the `sourceAccountId` and `payer` (just updated) fields may need to be persisted in the app from account linking. If I understand it correctly, the user needs to choose `sourceAccountId` in the app before initiating the transaction, and the `payer` data may need to be cached by the app (or perhaps could be stored by the server as well).

docs/assets/diagrams/transfer/03a-confirm-payee.puml
stevenwjy: Nope, just updated with the example values...

However, for some other fields, I currently still use placeholders as I'm not quite sure about the exact details and perhaps it may not be a big concern for the sequence diagram?

docs/README.md
```diff
@@ -2,6 +2,11 @@
 
 Documentation for the PISP demo server
 
+## Design
+
+There are two main scenarios that are handled by the PISP demo server:
+- [Transfer](./design/transfer.md)
```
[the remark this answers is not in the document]
stevenwjy: The linking will be added using [another PR](https://github.com/mojaloop/pisp-demo-server/pull/5).
Just updated the phrasing to make it less confusing...

docs/assets/diagrams/transfer/02-create-transaction.puml
```diff
@@ -0,0 +1,132 @@
...
+  ""    },""
+  ""    status: "PENDING"""
+  ""  }""
+  "")""
+end rnote
+
+Firebase -> Server: Notify event for new document
+return ""200 OK""
+
+Server -> Adapter: ""**GET /parties/{Type}/{ID}**""
```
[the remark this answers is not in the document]
stevenwjy: Updated

docs/assets/diagrams/transfer/02-create-transaction.puml
```diff
@@ -0,0 +1,132 @@
...
+box "Mobile Device" #Light
+  actor Alice
+  participant "PISP App" as App
+end box
+box "PISP" #Light
+  database Firebase
+  participant "PISP Server" as Server
+end box
+box "Mojaloop" #Light
+  participant "Third-party\nAPI Adapter" as Adapter
```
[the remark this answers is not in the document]
stevenwjy: I think based on my previous discussion with @lewisdaly, the PISP server will talk with `thirdparty-sdk-adapter` with will behave in a quite similar way with `sdk-scheme-adapter`, cmiiw.

docs/assets/diagrams/transfer/02-create-transaction.puml
```diff
@@ -0,0 +1,132 @@
...
+return ""**HTTP 202** (Accepted)""
+
+Adapter -> Adapter: Lookup data
+
+rnote left of Adapter #Light
+  ""**PUT /parties/{Type}/{ID}**""
+  ""{ ""
+  ""  "party": {""
+  ""    "partyIdInfo": {""
+  ""      "partyIdType": enum,""
```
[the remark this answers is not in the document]
stevenwjy: updated

docs/assets/diagrams/transfer/03a-confirm-payee.puml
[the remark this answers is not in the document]
stevenwjy: I think it's up to the app to decide (cc @MahidharBandaru). However, the app needs to pass in the value at this step because this will be the one that triggers calling `POST thirdpartyRequests/transactions`.

docs/assets/diagrams/transfer/03a-confirm-payee.puml
```diff
@@ -0,0 +1,127 @@
...
+  ""  "transactionRequests/<id>",""
+  ""  {""
+  ""    userId: 1234,""
+  ""    key: <uuid>""
+  ""  }""
+  "")""
+end rnote
+
+Server -> Firebase: Create mapping between\nuser info and transactionRequestId
+Server <-- Firebase: ""200 OK""
```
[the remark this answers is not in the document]
stevenwjy: Hmm... apparently it will result in something like this:
image

Probably because I still keep Adapter active to continue its execution in the next part of the sequence diagram.

docs/assets/diagrams/transfer/03b-confirm-payee.puml
```diff
@@ -0,0 +1,120 @@
+@startuml
+
+title Confirm Payee
```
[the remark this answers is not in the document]
stevenwjy: Okay, will update this

docs/assets/diagrams/transfer/02-create-transaction.puml
```diff
@@ -0,0 +1,132 @@
...
+Alice -> App: I want to make this transaction
+
+rnote right of App #Light 
+  ""createDoc(""
+  ""  "transactions/users/1234/<uuid>",""
+  ""  {""
+  ""    payee: {""
+  ""      phoneNumber: "+1-222-222-2222"""
+  ""    },""
+  ""    status: "PENDING"""
```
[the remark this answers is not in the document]
stevenwjy: Sorry, can I clarify what do you mean by "transaction notification"? are you referring to the result of party lookup?

> I think perhaps the server should be the only one permitted to update the status field, right?

I'm actually not sure about this. In the current design, after the user confirms the payee info (in diagram 3a), it will also change the status from `PENDING_PAYEE_CONFIRMATION` to `PAYEE_CONFIRMED`. 

However, if you want the server to be the only one who has write access to the Firebase, I actually think the app could talk directly to the server (instead of writing the message in Firebase which then notify the server). That way, the server will be the only one who always writes to the database and the app only needs to have a read permission to keep listening to all the changes.

This is actually also related to my concern when writing the sequence diagrams for linking. Because I don't think the app should let Firebase store signed challenge or auth token even for just a moment, and instead need to send its message directly to the server.

docs/assets/diagrams/transfer/02-create-transaction.puml
```diff
@@ -0,0 +1,124 @@
...
+  participant "PISP Server" as Server
+end box
+box "Mojaloop" #Light
+  participant "Third-party\nAPI Adapter" as Adapter
+end box
+
+Alice -> App: I want to make this transaction
+
+rnote right of App #Light 
+  ""createDoc(""
```
stevenwjy, line 24: To mention my previous discussion with @looilee, he suggested to separate the lookup to decouple it from the transaction. So perhaps the app can also listen to `lookups/users/1234` which stores all the result for party lookups performed and delete it once it is unused (cmiiw). That way, `createDoc("transactions/users/1234/<uuid>", ...)` will be performed only after the user confirms the payee and specify the transfer amount. In my opinion, it is also a good way to go since we may need to perform party lookup in the linking process as well. May I ask what do you think about this @jgeewax?

docs/assets/diagrams/transfer/03a-confirm-payee.puml
looilee: Please consider what happens if the app is uninstalled and reinstalled

MahidharBandaru: I was assuming this will be stored by the server. A list of linked accounts for an user of the app.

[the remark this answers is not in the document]
MahidharBandaru: Currently the user chooses where the money is coming from first.

stevenwjy:
Also, after talking with @lewisdaly about the current design, I just realised that `PISP Server` will need to talk with `Third-party SDK Adapter` instead of `Third-party API Adapter`. About the sdk adapter itself, it will work in a synchronous way. Hence, for example, instead of using `GET /parties/...` and wait for `PUT /parties/...` asynchronously like the way it is currently designed, we can just call `POST /parties/...` and then expect the result to be returned in the response body.

Currently, the plan is to keep using Firebase to sync between the server and mobile app, since the interaction between those two should be decoupled w.r.t. how the server talks with Mojaloop. 

To summarise my previous discussion with JJ, the main reason for us to stick with Firebase is to minimize the number of concurrent open connections. If I understand it correctly, it may take the sdk adapter some time (assume ~30s) to give a response for a particular request, since there are some processes going on internally to convert Mojaloop async behavior to synchronous. However, assuming we want to serve 1000 transactions per second, we may need to keep 30,000 open connections at the same time. With Firebase, the server only needs to keep one connection to serve all transactions.

Another problem that comes to mind is whether this will really solve the problem, because the communication between PISP server and the sdk scheme adapter may still become the bottleneck in terms of the number of concurrent connections. However, optimizing the infrastructure issue is probably not the current focus of the discussion...

cc @jgeewax @looilee

docs/assets/diagrams/transfer/03a-confirm-payee.puml
stevenwjy: Oh yeah, sorry I mean the server will also store the data, but the mobile app is also expected to cache the data locally.

lewisdaly:
As discussed in the call today, we are going to go ahead with the sdk/thirdparty scheme-adapter approach, as we are already building this as a part of the PISP work, and it makes sense to 'dog food' our own work here. We are having a call tomorrow to discuss the designs of the synchronous API that the scheme adapter will expose, and the required state machines.

Regarding the performance requirements, indeed we should add performance tests of the scheme adapter to our backlog, and consult the guys who wrote it to see if they have numbers already.

docs/assets/diagrams/transfer/03a-confirm-payee.puml
stevenwjy: Added a note about this in the documentation

docs/assets/diagrams/transfer/03a-confirm-payee.puml
stevenwjy: All updated with examples

docs/assets/diagrams/transfer/02-create-transaction.puml
```diff
@@ -0,0 +1,132 @@
...
+  ""  "transactions/users/1234/<uuid>",""
```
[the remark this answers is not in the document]
stevenwjy: Updated the syntax to perform queries to firebase

docs/assets/diagrams/transfer/02-create-transaction.puml
stevenwjy: In my recent update, I try to split the collections for transaction and lookup

docs/assets/diagrams/transfer/02-create-transaction.puml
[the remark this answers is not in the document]
stevenwjy: updated to communicate with Mojaloop switch

docs/assets/diagrams/transfer/02-create-transaction.puml
stevenwjy: Reverted to the original flow.

stevenwjy:
> 4. `firebase.database().ref('transactions').orderByKey('networkId').limitToFirst(1).once('value', (snap) => snap.key );`
> 
> This would mean we don't need a separate mapping between Mojaloop transaction request IDs and Mojapay's Firebase document IDs... Can you check that to make sure I'm not missing anything?

@jgeewax not sure what do you mean by `networkId` here, is it perhaps `transactionRequestId`? Besides, in the recent update, I removed the need to find mapping between `transactionRequestId` and Mojapay's Firebase document ID. Instead, the document name is the `transactionRequestId` itself. Any concerns with that?

The reason is because the PISP is responsible to decide on a unique `transactionRequestId` anyway before sending that information to the Mojaloop's switch. Hence, instead of performing repeated task on the server to generate an UUID and check whether it already exists in the database (even though the probability is super low to have a clash), I moved the checking to the client side. Now, the mobile app is expected to generate a random UUID, and create a document with that value under `transactions` collection. 

The advantage of doing it this way is that we won't need to set `transactionRequestId` as an index in the Firestore, which can probably make it more efficient since we already need to set `userId` as an index. However, the tradeoff is that the client may need to retry generation a unique `transactionRequestId` (with very low probability).

docs/assets/diagrams/transfer/03-confirm-payee.puml
```diff
@@ -0,0 +1,118 @@
...
+
+rnote right of App #Light
+  ""firebase.firestore()""
+  ""  .collection("transactions")""
+  ""  .doc("456")""
+  ""  .set({""
+  ""    sourceAccountId: "bob.fspb",""
+  ""    consentId: 123,""
+  ""    amount: {""
+  ""      amount: 20.00""
```
lewisdaly, line 31: I think this should technically be a string.

docs/assets/diagrams/transfer/03-confirm-payee.puml
```diff
@@ -0,0 +1,118 @@
...
+  ""    },""
+  ""    payer: {""
+  ""      personalInfo: {""
+  ""        complexName: {""
+  ""          firstName: "Bob",""
+  ""          lastName: "Beaver"""
+  ""        }""
+  ""      },""
+  ""      partyIdInfo: {""
+  ""        partyIdType: "OPAQUE",""
```
lewisdaly, line 42: I'm not sure this should be `OPAQUE` here. We were planning to use `OPAQUE` primarily for the account linking step when we know the dfsp, but _not_ the identifier type.

By this stage, I think we should have a proper identifier and type (e.g. `EMAIL`, `MSISDN` etc.) but the question remains: when do we actually get the "proper" mojaloop identifier?

docs/assets/diagrams/transfer/06-transaction-feedback.puml
```diff
@@ -0,0 +1,65 @@
...
+  participant "PISP Server" as Server
+end box
+box "Mojaloop" #Light
+  participant "Switch" as Mojaloop
+end box
+
+activate Mojaloop
+
+rnote left of Mojaloop #Light
+  ""**PUT /thirdpartyRequests/transactions/456**""
```
lewisdaly, line 24: This will be `PUT /transfers/{id}` now, based on more recent design discussions, as it is an exact forwarded request from the PayeeFSP to the switch

docs/assets/diagrams/transfer/03-confirm-payee.puml
stevenwjy: The original intention was to give the mobile app flexibility to pass in a floating point value and let the server handle the rounding since Mojaloop [does not allow trailing zeros] in the amount field... but yeah after thinking more about this, it may also better to handle the conversion in the mobile app to reduce the amount of work needed by the server.

Also, can I check whether you store it as a number or string in the mobile app @MahidharBandaru ?

docs/assets/diagrams/transfer/06-transaction-feedback.puml
stevenwjy: okay, can I check what's the format of the callback? is the current one still correct?

stevenwjy: @lewisdaly based on the [API Definition v1.0](https://docs.mojaloop.io/mojaloop-specification/documents/API%20Definition%20v1.0.html#6731-put-transfersid), `PUT /transfers/{ID}` does not contain any information about the `transactionRequestId`, which is needed to let PISP identify the transaction, any idea on how we could get that information?

lewisdaly: Yeah, well spotted.

So there's a couple ideas floating around for this:
1. The `transactionRequestId` is encoded in the `ilpPacket`, so worst case scenario, we can decode the `ilpPacket`
2. Mojaloop API v2.0 looks likely to include the `transactionRequestId` in the request body
3. We're also thinking that we can let the PISP specify the `transferId` ahead of time during the `POST thirdpartyRequests/transaction`

docs/assets/diagrams/transfer/03-confirm-payee.puml
stevenwjy: Do we get the proper identifier and type from account linking?

stevenwjy: Updated to use MSISDN for now

lewisdaly: Good question. I'm actually not sure. Initially we assumed we would have a _proper_ lookup in the beginning of the account linking, and so would have already saved the `partyIdType` and `partyIdentifier`. That has now changed (as you can tell).

Perhaps we can rely on the result (the `PUT /parties/OPAQUE/{id}`) from the `OPAQUE` account lookup at the start of account linking to have a concrete `partyIdType` (e.g. https://docs.mojaloop.io/mojaloop-specification/documents/API%20Definition%20v1.0.html#7411-party) in the request body.

@jgeewax  thoughts?

docs/assets/diagrams/transfer/06-transaction-feedback.puml
stevenwjy: > 1. The `transactionRequestId` is encoded in the `ilpPacket`, so worst case scenario, we can decode the `ilpPacket`

But `PUT /transfers/{ID}` also does not contain `ilpPacket` right? Or perhaps is it included inside the `fulfilment`?

> 2. Mojaloop API v2.0 looks likely to include the `transactionRequestId` in the request body
> 3. We're also thinking that we can let the PISP specify the `transferId` ahead of time during the `POST thirdpartyRequests/transaction`

If we are going to include both `transactionId` and `transactionRequestId` everywhere, may I ask about why do we need to separate those values? Wouldn't it be better to have a single id that identifies the whole transaction process?

docs/assets/diagrams/transfer/03-confirm-payee.puml
stevenwjy: okay, will create a ticket for it and update this comment

docs/assets/diagrams/transfer/06-transaction-feedback.puml
lewisdaly: >But PUT /transfers/{ID} also does not contain ilpPacket right? Or perhaps is it included inside the fulfilment?

Yeah you're right, I was getting confused with `POST /transfers`

>If we are going to include both transactionId and transactionRequestId everywhere, may I ask about why do we need to separate those values? Wouldn't it be better to have a single id that identifies the whole transaction process?

Sorry I meant just `transactionId`, not `transactionRequestId`.

stevenwjy: > Sorry I meant just `transactionId`, not `transactionRequestId`.

Sorry, could you please reiterate on this? If I understand it correctly, PISP needs the `transactionRequestId` to identify the transfer right? And unless it is the one who decides on the `transactionId`, I assume that getting the value won't be useful? However, my concern here is that, if PISP is the one that decides on both values, why do we need to differentiate them?

lewisdaly: Yeah no worries. You've uncovered a part of the design that has more _hand waving_ than the rest, and we aren't 100% solid on what this will look like just yet.

I think we're leaning on the idea that a PISP can specify the `transferId` and `transactionId` ahead of time in the `POST /thirdpartyRequests/transactions` request. That way, it _can_ understand and identify incoming `PUT /transfers/{id}` requests, since it was one who specified the `transferId`.

> However, my concern here is that, if PISP is the one that decides on both values, why do we need to differentiate them? 

You're right - I guess we don't strictly need to differentiate them, but remember that they have different purposes. A transaction (and hence a `transactionId`) can live outside the scope of a transaction request. 

If `transactionId` is added to the `Transfer` object, it would be a _required_ field (I think), because a transfer must always be within a bigger transaction (even if the transaction only contains 1 transfer).

A `TransactionRequest` (and for our purposes, `ThirdpartyTransactionRequest`) may not result in the creation of a `Transfer` or `Transaction`.

Additionally, not every `Transfer` is the result of a `TransactionRequest`, so if we added the `transactionRequestId` to the `Transfer` Object (which I don't think is being discussed at the moment, but I could be wrong), it would be optional, much like in the `Quotes` Object. (see [post-quotes](https://docs.mojaloop.io/mojaloop-specification/documents/API%20Definition%20v1.0.html#6522-post-quotes))

stevenwjy: Ahh I see, okay got it, thanks for the robust explanation!

Besides, I just realised something with the current transfer flow design...

If we go with that option, the database for PISP will need to have an additional index, which is for the `transactionId`, because it needs to handle `POST /authorizations`  with only `transactionRequestId` being provided, but also need to handle `PUT /transfers/{ID}` with only `transactionId` being provided.

Is it safe to assume that we can set `transactionId` equal to `transactionRequestId` to simplify the indexing?

docs/assets/diagrams/transfer/02-create-transaction.puml
```diff
@@ -0,0 +1,126 @@
...
+  ""        partyIdType: "MSISDN",""
+  ""        partyIdentifier: "+1-222-222-2222",""
+  ""        fspId: "fspa"""
+  ""      }""
+  ""    },""
+  ""    status: "PENDING_PAYEE_CONFIRMATION",""
+  ""  }, { merge: true })""
+end rnote
+
+Server -> Firebase ++: Update transaction status
```
stevenwjy, line 108: Btw, I just realized one remaining issue here...

Upon receiving `PUT /parties/MSISDN/+1-222-222-2222`, the server does not know about the `transactionRequestId` associated with that request, which in this case is equal to `456`.

One way that I can think to solve this problem is by storing a temporary document in Firebase to track the list of `transactionRequestId` that are waiting to receive callback with that party identifier.

Alternatively, we can also imitate how `sdk-scheme-adapter` handles this issue, which is by using redis pub/sub (assuming we want to launch multiple instances of this server; otherwise, we can just keep everything in memory). However, it will require the server to also keep the subscription channels open while waiting for reply.

Any thoughts on this? @jgeewax @looilee @lewisdaly

docs/assets/diagrams/transfer/03-confirm-payee.puml
```diff
@@ -0,0 +1,118 @@
...
+  ""  "sourceAccountId": "bob.fspb",""
+  ""  "consentId": "789",""
+  ""  "payee": {""
+  ""    "partyIdInfo": {""
+  ""      "partyIdType": "MSISDN",""
+  ""      "partyIdentifier": "+1-222-222-2222",""
+  ""      "fspId": "fspa"""
+  ""    }""
+  ""  },""
+  ""  "payer": {""
```
looilee, line 80: Does the app need to specify this or should/can it be done by the server?

docs/assets/diagrams/transfer/03-confirm-payee.puml
```diff
@@ -0,0 +1,118 @@
...
+  ""  }, { merge: true })""
+end rnote
+
+App -> Firebase ++: Create a new transaction document
+Firebase <-- App: ""200 OK""
+
+rnote right of Firebase #Light
+  ""event(""
+  ""  "document.create",""
+  ""  "transactions/456",""
```
looilee, line 56: this is not created under transactions/users/*, is this intended?

docs/assets/diagrams/transfer/03-confirm-payee.puml
stevenwjy: In the current design, the app need to specify the payer field, because otherwise the server would need to make a query to Firebase for every transaction document. I think the data will be stored in Firebase after once an account has been linked, but the app is also expected to cache the value (e.g, after linking an account or when the app is opened for the first time).

docs/assets/diagrams/transfer/03-confirm-payee.puml
stevenwjy: Yes, this is based on [JJ's suggestion](https://github.com/mojaloop/pisp-demo-server/pull/4#pullrequestreview-443556058) to store everything under `transactions` collection so that the server could just listen to `transactions`. 

And then, the user will listen to all documents with field `userId` matching the id of the user based on Firebase auth.

docs/assets/diagrams/transfer/03-confirm-payee.puml
looilee: So what happens if the app is uninstalled and installed back again?

Also can you elaborate on "I think the data will be stored in Firebase after once an account has been linked"

stevenwjy: When the app runs for the first time, it will go through the following process:
1. Query to the server about what are the bank accounts that have been linked with a particular PISP account.
2. All the linked bank accounts are shown as "session expired" since we know that they have been linked previously but there's no consent to make transactions stored in the device.
3. If a user wants to make a transaction, he/she must perform an account linking again to get `consentId`.
4. Upon performing the account linking, the app will get the party information for each linked bank account, which then will be put in the `payer` field when initiating a transaction.

looilee: For point 1, how does the PISP server know what are the accounts has been linked? Do we persist that in firebase? I think this is related to your last point "that there could be a document to store what are the list of bank accounts that have been linked with a particular PISP account". You might want to make that decision sooner than later.

stevenwjy: Yeah, I think we can persist that ... but haven't really thought about it since we haven't discussed the linking process in detail...

For now, can we assume that if an app is uninstalled and then installed again, it will be considered as a fresh start (no account has been linked)? This is based on the assumption that all the public-private key pairs will be gone if the app is uninstalled, so we always have to perform an account linking before making any transaction.

looilee: Sure. But it doesn't sound like we have enough information to do what point 2 is saying as in displaying "session expired". I think it's fine just to go back to a blank slate for now just to make things easier (and manage less state).

docs/assets/diagrams/transfer/06-transaction-feedback.puml
stevenwjy: After looking back at the sequence diagram, I just realized that `POST /authorizations` actually contains both `transactionRequestId` and `transactionId`, which essentially solves the issue that we discussed here 😂 ... Hence, it's fine to let DFSP to decide on the `transactionId`. The PISP can create a mapping between `transactionRequestId` and `transactionId` upon receiving `POST /authorizations`.

In the current design, I still set `transactionRequestId` as a document ID in Firebase to ensure uniqueness, and just believe that `transactionId` will be unique as well... but in case `transactionId` is not unique, we will just end up updating multiple transaction documents and I think its beyond the control of the PISP...

docs/assets/diagrams/transfer/02-create-transaction.puml
stevenwjy: To resolve the current issue, I added a temporary field `partyQuery` which will be removed upon receiving a party lookup result... Not sure whether there's a better way to do that...

docs/assets/diagrams/transfer/03-confirm-payee.puml
stevenwjy: Yeah, I think we can discuss about this further when deciding on the linking flow

stevenwjy:
Btw, just updated the sequence diagram and went through it one more time... I think the logic seems to work for now :)

docs/assets/diagrams/transfer/02-create-transaction.puml
looilee: How about passing "456" as a query param of like:

GET /parties/MSISDN/+1-222-222-2222?transactionRequestId=456

stevenwjy: I think the problem is about how to identify the `transactionRequestId` upon receiving `PUT /parties/{Type}/{ID}`. Unless we store the mapping somewhere, I don't think passing a `transactionRequestId` in the query param will have any effect (?)

looilee: Sorry I thought I commented but I think I forgot to click on submit. You can have the Mojaloop switch return the same `transactionRequestId` it received in the PUT request, would that not work? This requires some small changes on the Mojaloop switch side.

stevenwjy: > Couldn't there be multiple transactions with the same partyQuery ?

Sorry, I think i missed replying to this... Yeah, because at the moment we couldn't differentiate between two different `PUT /parties/...` request



> You can have the Mojaloop switch return the same `transactionRequestId` it received by putting it in the PUT request, would that not work? This requires some small changes on the Mojaloop switch side.

I'm actually not sure about it... would it be possible to do something like this @lewisdaly ?

lewisdaly: Hmm, that's not going to be possible, but perhaps the issue here is that you are trying to tie together a party lookup and a transaction request before the transaction request even exists.

2 other ideas:
1. You add indexes to the `transactions` collection on `partyIdType` and `partyIdentifier`, and then perform a filter to lookup the `transactions.id` later on
2. You could make a temporary collection, something like `partyLookups` that stores the `transaction.id` and callback state (like you said above)

docs/assets/diagrams/transfer/06-transaction-feedback.puml
lewisdaly: Ah ok. I forgot about how cloud firestore works in relation to indicies.

>Hence, it's fine to let DFSP to decide on the transactionId. The PISP can create a mapping between transactionRequestId and transactionId upon receiving POST /authorizations.

Sure, as long as you can build that relationship, I think we're fine, and can proceed for now. But I do think that the PISP will be setting the `transactionId`. And we _can_ assume a `transactionId` is unique.

> Besides, @lewisdaly, can I check whether the id in PUT /transfers/{ID} is equal to transactionId? Or perhaps there could be a different transferId? For now, I assume that they are different, but there's a transactionId in the request body.

Yeah I think we underestimated the challenges for Mojaloop Clients with respect to this callback. All I can say for now is watch this space as we still have unresolved design issues here. For now however, maybe we can assume that the `PUT /transfers/{ID}` will _contain_ a `transactionId` in the body (which it doesn't at the moment).

docs/assets/diagrams/transfer/02-create-transaction.puml
stevenwjy: Oh okay, I think currently what I'm doing with `partyQuery` temporary field is basically the same with your first suggestion.

docs/assets/diagrams/transfer/05-authorize-transaction.puml
```diff
@@ -0,0 +1,80 @@
...
+  ""firebase.firestore()""
+  ""  .collection("transactions")""
+  ""  .doc("abc123")""
+  ""  .set({""
+  ""    authenticationInfo: {""
+  ""      authentication: "U2F",""
+  ""      authenticationValue: {""
+  ""        pinValue: "base64(xxx)",""
+  ""      }""
+  ""    },""
```
jgeewax, line 33: Please think about what the optimal representation here is. It sounds to me like this is only represented this way because it's the way Mojaloop does it. But we DO NOT want to be a simple proxy server between a phone and Mojaloop. We need to be an ABSTRACTION layer.

stevenwjy: Updated the field to the following object representation:
```
authentication: {
  type: "U2F",
  value: "base64(xxx)"
}
```

The rationale behind this change:
- The `type` field is to support flexibility in case we would like to support other authentication methods in the future (e.g., OTP, Web)
- The `value` field is a generic field that can contain any necessary value to perform the authentication (e.g., FIDO signature, OTP code, short-lived token).

Any thoughts on the new authentication fields?

stevenwjy:
Just want to put a note here, the reason for using `set( ..., { merge: true })` instead of `update( ... )` for the firebase operation is because of the differences in handling nested fields.

Suppose we have the following document:
```
{
  "a": {
    "b1": "test1"
  }
}
```

If we perform the following operation:
```
docRef.set({
  "a": {
    "b2": "test2"
  }
}, { merge: true })
```

The resulting document is:
```
{
  "a": {
    "b1": "test1",
    "b2": "test2"
  }
}
```

However, if we use:
```
docRef.update({
  "a": {
    "b2": "test2"
  }
})
```


The resulting document is:
```
{
  "a": {
    "b2": "test2"
  }
}
```

[Reference](https://stackoverflow.com/questions/46597327/difference-between-set-with-merge-true-and-update)

PR #8, opened 2020-07-16
Add implementation for payee confirmation and auth prompt in the transfer flow

stevenwjy:
This PR depends on the functionalities that are implemented in #7. 

- [x] Handle update from Firebase that represents payee confirmation for a transfer
- [x] Send `POST /thirdpartyRequests/transactions` to Mojaloop (mocked)
- [x] Handle callback from Mojaloop that prompts user for authorization, i.e., `POST /authorizations`
- [x] Update Firebase to contain all necessary data to prompt user for authorization in the mobile app.


src/shared/ml-thirdparty-client/models/core/transactions.ts
```diff
@@ -35,3 +54,137 @@ export enum Currency {
...
+
+  /**
+   * Who is initiating the transaction - Payer or Payee.
+   */
+  initiator: string
+
+  /**
+   * Consumer, agent, business, …
+   */
+  intiiatorType: string
```
looilee, line 178: typo

src/shared/ml-thirdparty-client/models/core/transactions.ts
```diff
@@ -35,3 +54,137 @@ export enum Currency {
...
+ * Quote of a transaction.
+ */
+export interface Quote {
+  /**
+   * The amount of money that the Payee FSP should receive.
+   */
+  transferAmount: Money
+
+  /**
+   * The amount of Money that the Payee should receive in the end-to-end transaction.
```
looilee, line 99: Is payeeReceiveAmount equal to transferAmount - payeeFspFee? If so, would be helpful to describe that explicitly here.

src/shared/ml-thirdparty-simulator/factories/__mocks__/authorization.ts
```diff
@@ -0,0 +1,51 @@
...
+ - Name Surname <name.surname@gatesfoundation.com>
+
+ * Google
+ - Steven Wijaya <stevenwjy@google.com>
+ --------------
+ ******/
+
+import {
+  ThirdPartyTransactionRequest,
+  AuthorizationsPostRequest,
```
looilee, line 28: why does this have a trailing comma....

src/shared/ml-thirdparty-simulator/factories/__mocks__/authorization.ts
```diff
@@ -0,0 +1,51 @@
...
+} from '~/shared/ml-thirdparty-client/models/openapi'
+
+import {
+  AuthenticationType
```
looilee, line 32: but not this?

test/unit/server/handlers/firestore/transactions.test.ts
```diff
@@ -43,6 +46,64 @@ jest.mock('uuid', () => ({
...
 
+// Mock utils to consistently return the provided value.
+jest.mock('~/lib/utils', () => ({
+  getTomorrowsDate: jest.fn().mockImplementation(() => {
+    return new Date(100)
+  })
+}))
+
+// Define mock data to perform transaction request
+function mockTransactionRequestData(): Transaction {
```
looilee, line 57: I think is a stub rather than mock.

test/unit/server/handlers/firestore/transactions.test.ts
```diff
@@ -43,6 +46,64 @@ jest.mock('uuid', () => ({
...
+          lastName: 'Alpaca',
+        },
+      },
+    },
+    status: Status.PENDING_PAYEE_CONFIRMATION,
+  }
+}
+
+// Define mock consent data related to the transaction
+function mockConsentData(): Consent {
```
looilee, line 86: ditto

src/shared/ml-thirdparty-client/models/core/transactions.ts
stevenwjy: Oops... sorry didn't notice this one, fixed :)

src/shared/ml-thirdparty-simulator/factories/__mocks__/authorization.ts
stevenwjy: this is the correct one, fixed the inconsistency on the other parts of the code

src/shared/ml-thirdparty-simulator/factories/__mocks__/authorization.ts
stevenwjy: fixed

test/unit/server/handlers/firestore/transactions.test.ts
stevenwjy: Okay, fixed

test/unit/server/handlers/firestore/transactions.test.ts
stevenwjy: fixed

src/shared/ml-thirdparty-client/models/core/transactions.ts
stevenwjy: For this part, I actually just copied the description from the Mojaloop OpenAPI definition. There are quite a lot of things that could affect the `payeeReceiveAmount`, which are the commissions and fees from both Payee and Payer FSPs. If we are about to write the calculation here, I think it would be quite out of scope since the readers will also need to understand how the commissions and fees could cover for each other, how the Payee FSP may hide the fee and get commission from the Payer FSP, etc.  Hence, I feel like it would be better to keep things simple by just writing the comment as it is... What do you think about this?

looilee: Yes you can just describe it in general, something like payeeReceiveAmount is usually transferAmount less commission and fees. No need to write an essay about it.

stevenwjy: Okay, added a comment about it

PR #9, opened 2020-07-16
Add implementation for authorization and handling transaction feedback in the transfer flow

stevenwjy:
This PR depends on the functionalities that are implemented in #8 

- [X] Handle update from Firebase that represents payer authorization for a transaction
- [X] Send `PUT /authorizations/{ID}` request to Mojaloop (mocked)
- [X] Handle callback from Mojaloop that contains the transaction status
- [X] Update Firebase based on the transaction status

While waiting for #8 to get merged, [here](https://github.com/stevenwjy/pisp-demo-server/pull/2) is a tracking pr to compare the difference between pt3 and pt2.


src/shared/ml-thirdparty-client/models/openapi/authorizations.ts
```diff
@@ -63,3 +63,15 @@ export interface AuthorizationsPostRequest {
    */
   quote: Quote
 }
+
+export interface AuthorizationsPutIdRequest {
+  /**
+   * Authentication value provided by payer if entered.
+   */
+  authenticationInfo?: AuthenticationInfo
```
kkzeng, line 71: why is authenticationInfo optional?

stevenwjy: yeah because the `responseType` could be equal to `REJECTED` which means the user does not provide any authentication

PR #10, opened 2020-08-03
Update the sequence diagrams for transfer flow

stevenwjy:
- Update the usage of `authentication.field` in Firebase 
    - For authentication prompt, the server will update `authentication.type` instead of `authenticationType` 
    - For authorization, the app does not need to update the `authentication.type` value
- Change to use `update` function to update Firebase instead of `set` for operations that do not need deep merging. 
- Update server to listen only to uncompleted transactions to improve performance.


MahidharBandaru:
LGTM :+1:

PR #13, opened 2020-08-13
Feature/401: Setup Sdk-Standard-Components request objects

ramanmangla:
This is in reference to ticket #401 .

- Setup `sdk-standard-components` config for requests
- Add `ambient.d.ts` definitions for `MojaloopRequests`
- Refactor `MojaloopClient` custom plugin to integrate sdk request objects
- Add `sdk-standard-components` config properties to 'Convict' schema


src/shared/ml-thirdparty-client/config.ts
```diff
@@ -0,0 +1,34 @@
...
+
+const config = {
+  PARTICIPANT_ID: 'pisp',
+  ALS_ENDPOINT: 'account-lookup-service:4002',
+  THIRDPARTY_REQUEST_ENDPOINT: 'thirdparty-api-adapter:3008',
+  TRANSACTION_REQUEST_ENDPOINT: 'transaction-request-service:4003',
+  PEER_ENDPOINT: '172.17.0.2:3001',
+}
+
+export default config
```
stevenwjy, line 34: Can you please update and move the config schema that is written in the `index.ts` file instead of creating a new one here? Besides, I believe that we shouldn't just hardcode the values like this because `ml-thirdparty-client` is intended to be a configurable plugin. It would make more sense to treat the current hardcoded values as default config instead.

src/shared/ml-thirdparty-client/requests.ts
```diff
@@ -0,0 +1,62 @@
...
+
+ * Google
+ - Raman Mangla <ramanmangla@google.com>
+ --------------
+ ******/
+
+import config from './config'
+import Logger, {
+  ThirdpartyRequests,
+  MojaloopRequests,
```
stevenwjy, line 34:
```suggestion
   MojaloopRequests,
```

src/shared/ml-thirdparty-client/requests.ts
```diff
@@ -0,0 +1,62 @@
...
+      },
+    },
+  },
+  peerEndpoint: config.PEER_ENDPOINT,
+  alsEndpoint: config.ALS_ENDPOINT,
+  thirdpartyRequestsEndpoint: config.THIRDPARTY_REQUEST_ENDPOINT,
+  transactionRequestsEndpoint: config.TRANSACTION_REQUEST_ENDPOINT,
+}
+
+export const thirdPartyRequest: ThirdpartyRequests = new ThirdpartyRequests(
```
stevenwjy, line 56: I think we need to be consistent by writing it as `thirdparty` instead of `thirdParty`

src/shared/ml-thirdparty-client/requests.ts
```diff
@@ -0,0 +1,62 @@
...
+  configRequest
+)
+
+export const mojaloopRequest: MojaloopRequests = new MojaloopRequests(
+  configRequest
+)
```
stevenwjy, line 62: Do you think this will introduce a problem during testing if we use `ThirdpartyRequests` and `MojaloopRequests` everywhere? After thinking about this again, I think it would be easier if we create another wrapper that only has functions that we need instead of all possible functions provided by `ThirdpartyRequests` and `MojaloopRequests`. That way, we can easily use dependency injection for the testing.

ramanmangla: But again, that is why we have a Mojaloop client wrapper class: to abstract these methods. Otherwise, there is no point of having a Mojaloop client class.

stevenwjy: hmm okay, that's fair

src/shared/ml-thirdparty-client/index.ts
```diff
@@ -20,56 +20,47 @@
...
+  mojaloop: MojaloopRequests
 
-  public constructor(config?: Client.Config) {
-    if (config) {
-      this.config = { ...defaultConfig, ...config }
-    } else {
-      this.config = defaultConfig
-    }
+  public constructor() {
+    // if (config) {
```
stevenwjy, line 51: why do you remove the config from the constructor? I think we still need to pass in the config from the constructor since we want to model this as a plugin, so it does not read from a config file directy

src/shared/ml-thirdparty-client/config.ts
```diff
@@ -0,0 +1,86 @@
...
+
+// namespace Client {
+/**
+ * An interface definition for the configuration needed to setup the
+ * Mojaloop client.
+ */
+
+import Convict from 'convict'
+
+export interface clientConfigModel {
```
stevenwjy, line 34: `ClientConfig` is a better naming I guess if you don't want to put it under a namespace

src/shared/ml-thirdparty-client/config.ts
```diff
@@ -0,0 +1,86 @@
...
+  MOJALOOP_URL: string
```
stevenwjy, line 35: can we make this consistent with other naming using camel case?

src/shared/ml-thirdparty-client/index.ts
akapur99: yeah this constructor feels a little confusing

src/shared/ml-thirdparty-client/requests.ts
akapur99: do we even need the client class though? instead of object methods why not just make it a library of request functions which uses the TPR and MJR objects and calls everything accordingly? 
I don't see the need for an additional layer of abstraction here - what is the benefit?

stevenwjy: Yup, the `Client` class is doing exactly that at this time

PR #15, opened 2020-08-21
Cleanup code

stevenwjy:
This PR aims to clean up various things in the PISP demo server code, while at the same time trying to not break any of the existing functionalities. There is only one goal here: make the code looks more neat! :-) 

Here are the changes included in this PR:
- Remove open API endpoints that were defined for the communication with the mobile app. Things have changed in the past 1-2 months, and we no longer need those endpoints since the server will communicate with Firebase instead of the app directly. Hence, this PR removes all of the endpoints except for the health check. 
- Move the internal plugin for `ml-thirdparty-client` and `ml-thirdparty-simulator` to the shared folder (`hapi-plugin.ts` file). The main reason is to avoid code duplication when setting up the configuration.
- Avoid the confusion between `Config` and `Options`. All of the plugins now use a single term, which is `Options`.
- Add more comments to some parts of the code to improve readability


eoln (approved):
lgtm

PR #17, opened 2020-08-21
Add write-ups for the linking and transfer flows

stevenwjy:
This PR adds some more explanation to the documentation for both the transfer and linking flows as there were only sequence diagrams. Hopefully, this will improve the readability for people who want to discover more about how both important phases to connect with Mojaloop.


eoln (approved):
lgtm

PR #20, opened 2020-08-26
Update readme to include desc and setup guide.

kkzeng:
Make it easy to setup the server for other developers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,8 @@ A server used to communicate with the Mojaloop network to facilitate account lin
 Follow the instructions outlined [here](https://firebase.google.com/docs/admin/setup) to obtain a JSON file containing your secret key.
 Rename it to be `serviceAccountKey.json` and put it in `{project-directory}/secret/` so that the path to the file is: `{project-directory}/secret/serviceAccountKey`.
 
+Note that this path is just the default path but can be configured (see [Configurations section](#configurations).
+
 The contents of the JSON file should look like
 ```
 {
@@ -34,6 +36,7 @@ EXPERIMENTAL_MODE=on
 EXPERIMENTAL_DELAY=5000
 ```
 These will enable the Mojaloop simulator until the relevant features on the real Mojaloop network are implemented.
+To see other environment variables that you can use to customize the server configuration (see [Configurations section](#configurations).
 
 
 3. Mock Consent Objects
@@ -61,13 +64,17 @@ You need to create a collection in your Firestore named "consents" and then crea
 
 Note: The nested structure can be replicated by using the Map data type in Firestore.
 
-You can use any info for the partyIdentifier and names but the consentId must be 555 for now - this is hardcoded in the serverside code.
+You can use any info for the partyIdentifier and names but the consentId must be 555 for now - this is hardcoded in the mobile app so the consentId must match.
 
 (You can copy the example exactly and it will work)
 
-This represents the consent that the user has given to PISP app to be able to access their account information. When account linking is fully functional, this step will no longer need to be performed since we will get actual consent objects in the collection.
+This represents the consent that the user has given to PISP app to be able to access their account information. When account linking is fully functional, this step will no longer need to be performed since we will get actual consent objects in the collection from the account linking process.
 
 # Starting the server
 After all of the steps are done. Type `npm run start` in the command line in the project directory.
 
 Go to the PISP demo app and try to send money to a payee. You should see that the Firestore collection "transactions" has a new document and that the document status is changing as the server performs each step in the transaction sequence.
+
+# Configurations
+
+Take a look at `src/lib/config` to see all the different aspects of the server that you can configure.
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ A server used to communicate with the Mojaloop network to facilitate account lin
 Follow the instructions outlined [here](https://firebase.google.com/docs/admin/setup) to obtain a JSON file containing your secret key.
 Rename it to be `serviceAccountKey.json` and put it in `{project-directory}/secret/` so that the path to the file is: `{project-directory}/secret/serviceAccountKey`.
 
-Note that this path is just the default path but can be configured (see [Configurations section](#configurations).
+Note that this path is just the default path but can be configured (see [Configurations section](#configurations)).
 
 The contents of the JSON file should look like
 ```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1 +1,72 @@
 # PISP Demo Server (Work in Progress)
+
+A server used to communicate with the Mojaloop network to facilitate account linking and transactions for the PISP demo app (see: https://github.com/mojaloop/pisp-demo-app-flutter).
```
stevenwjy, line 3:
```suggestion
A server used to communicate with the Mojaloop network to facilitate account linking and transactions for the [PISP demo app](https://github.com/mojaloop/pisp-demo-app-flutter).
```

README.md
```diff
@@ -1 +1,72 @@
...
+
+# Quick Setup Guide
+
+1. Firebase SDK Admin Key
+
+Follow the instructions outlined here: https://firebase.google.com/docs/admin/setup to obtain a .json file containing your secret key.
```
stevenwjy, line 9:
```suggestion
Follow the instructions outlined [here](https://firebase.google.com/docs/admin/setup) to obtain a JSON file containing your secret key.
```

README.md
```diff
@@ -1 +1,72 @@
...
+Rename it to be `serviceAccountKey.json` and put it in `{project-directory}/secret/` so that the path to the file is: `{project-directory}/secret/serviceAccountKey`.
```
stevenwjy, line 10: Actually for this one, it could be customised from the config. Perhaps can mention that people can just put the path in the env variable or config file, but the default one is `{project-directory}/secret/serviceAccountKey`?

README.md
```diff
@@ -1 +1,72 @@
...
+  "token_uri": "https://oauth2.googleapis.com/token",
+  "auth_provider_x509_cert_url": "https://www.googleapis.com/oauth2/v1/certs",
+  "client_x509_cert_url": "https://www.googleapis.com/robot/v1/metadata/x509/firebase-adminsdk.....iam.gserviceaccount.com"
+}
+```
+
+2. Environment Variables
+
+Create a `.env` file in your local environment.
+Put these lines into that file:
```
stevenwjy, line 31: I think we can mention that people can take a look at the config definition under `src/lib/config` to find out more about other configurations that could be setup.

README.md
```diff
@@ -1 +1,72 @@
...
+```
+These will enable the Mojaloop simulator until the relevant features on the real Mojaloop network are implemented.
+
+
+3. Mock Consent Objects
+
+You need to create a collection in your Firestore named "consents" and then create an object in that collection like so: 
+
+```
+consentId: "555"
```
stevenwjy, line 44:
```suggestion
{
  "consentId": "555"
```

The document is a single JSON structure

README.md
```diff
@@ -1 +1,72 @@
...
+         "firstName":"Bob",
+         "lastName":"Beaver"
+      }
+   }
+}
+```
+
+Note: The nested structure can be replicated by using the Map data type in Firestore.
+
+You can use any info for the partyIdentifier and names but the consentId must be 555 for now - this is hardcoded in the serverside code.
```
stevenwjy, line 63: It's hardcoded in the mobile app not the server, actually its just the stub data which is currently used since we haven't implemented the account linking. Maybe can mention that this will change soon once account linking is there?

stevenwjy (changes requested):

README.md
kkzeng: I will include it as a separate section under Configuration or something since the idea is to quickly setup the server.

README.md
kkzeng: I will correct it to say mobile app. I already mentioned below that this will change.

README.md
```diff
@@ -1 +1,80 @@
...
+
+
+3. Mock Consent Objects
+
+You need to create a collection in your Firestore named "consents" and then create an object in that collection like so: 
+
+```
+{
+  "consentId": "555"
+"party":{
```
stevenwjy, line 49: I think you also need to fix this based on the previous line

README.md
```diff
@@ -1 +1,80 @@
...
+         "firstName":"Bob",
+         "lastName":"Beaver"
+      }
+   }
+}
+```
+
+Note: The nested structure can be replicated by using the Map data type in Firestore.
+
+You can use any info for the partyIdentifier and names but the consentId must be 555 for now - this is hardcoded in the mobile app so the consentId must match.
```
stevenwjy, line 67: actually if you take a look at the `StubAccountRepository` in the mobile app, there are two accounts there with different consentId, do you perhaps want to write both?

README.md
```diff
@@ -1 +1,80 @@
...
+2. Environment Variables
+
+Create a `.env` file in your local environment.
+Put these lines into that file:
+```
+EXPERIMENTAL_MODE=on
+EXPERIMENTAL_DELAY=5000
+```
+These will enable the Mojaloop simulator until the relevant features on the real Mojaloop network are implemented.
+To see other environment variables that you can use to customize the server configuration (see [Configurations section](#configurations).
```
stevenwjy, line 39: missing closing bracket? same with the other one

README.md
kkzeng: yep, fixed oops

README.md
kkzeng: 985 right, ok included

README.md
```diff
@@ -1 +1,81 @@
...
+This represents the consent that the user has given to PISP app to be able to access their account information. When account linking is fully functional, this step will no longer need to be performed since we will get actual consent objects in the collection from the account linking process.
+
+# Starting the server
+After all of the steps are done. Type `npm run start` in the command line in the project directory.
+
+Go to the PISP demo app and try to send money to a payee. You should see that the Firestore collection "transactions" has a new document and that the document status is changing as the server performs each step in the transaction sequence.
+
+# Configurations
+
+Take a look at `src/lib/config` to see all the different aspects of the server that you can configure.
```
stevenwjy, line 81: maybe can give link to the file here?

README.md
```diff
@@ -1 +1,81 @@
...
+}
+```
+
+2. Environment Variables
+
+Create a `.env` file in your local environment.
+Put these lines into that file:
+```
+EXPERIMENTAL_MODE=on
+EXPERIMENTAL_DELAY=5000
```
stevenwjy, line 36: probably should change this into a smaller number :o, the delay is 5 seconds. Or perhaps you can also remove this since the default is 1 or 2 secs if I remember correctly

stevenwjy (approved):
lgtm 👍
just some extra nit comments

PR #22, opened 2020-08-27
Transfer + Linking/Unlinking Communication with Mojaloop

akapur99:
Addresses : 
* mojaloop/project#1649
* mojaloop/project#1628

**Tasks**:
- [x] Implement client library linking/unlinking related methods
- [x] Implement client library transfer related methods
- [x] Implement corresponding methods in client simulator
- [x] Create ConsentFactory for mocking incoming requests
- [x] Units tests for all of the above


lewisdaly:
Tests are failing - please use a draft PR if this isn't ready

src/shared/ml-thirdparty-client/index.ts
```diff
@@ -157,4 +158,123 @@ export class Client {
 
     // TODO: Implement communication with Mojaloop.
   }
+
+  /**
+   * Gets a list of PISP/DFSP participants
+   */
+  public async getParticipants(): Promise<void> {
+    if (this.simulator) {
```
jgeewax, line 166: Is this the standard for other Mojaloop code? This seems... a bit scary.

src/shared/ml-thirdparty-client/index.ts
```diff
@@ -157,4 +158,123 @@ export class Client {
...
+      // If the client is configured with a simulator, then it will not
+      // communicate with Mojaloop directly. Instead, it will only generate
+      // a random response that is injected to the internal routes.
+      return this.simulator.postConsentRequests(requestBody)
+    }
+    await this.thirdpartyRequests.postConsentRequests(
+      requestBody,
+      destParticipantId
+    )
```
jgeewax, line 194: I think generally this is something best handled by dependency injection. E.g., you provide a library (such as `this.thirdpartyRequests`) and then do the same actions on it...

So ultimately, this code would be something like

```typescript
return this._requestClass.postConsentRequests(requestBody, destParticipantId);
```

And the way you'd "turn on the simulator" would be by setting `_requestClass` to what you currently have in `this.simulator`...

src/shared/ml-thirdparty-client/index.ts
```diff
@@ -157,4 +158,123 @@ export class Client {
...
+    requestBody: SDKStandardComponents.PutConsentRequestsRequest,
+    destParticipantId: string
+  ): Promise<void> {
+    if (this.simulator) {
+      // If the client is configured with a simulator, then it will not
+      // communicate with Mojaloop directly. Instead, it will only generate
+      // a random response that is injected to the internal routes.
+      return this.simulator.putConsentRequests(consentRequestId, requestBody)
+    }
+    await this.thirdpartyRequests.putConsentRequests(
```
jgeewax, line 215: Why await here? Isn't that just blocking and then returning nothing? We shouldn't await unless we need the result for something else. If it's the end of a function, we should return the promise and allow the caller to decide whether to await or not...

src/shared/ml-thirdparty-client/index.ts
```diff
@@ -157,4 +158,123 @@ export class Client {
...
+   *
+   * @param consentId     identifier of consent as defined by Mojaloop API.
+   * @param requestBody         an object to authenticate consent as defined by the Mojaloop API.
+   * @param destParticipantId   ID of destination - to be used when sending request
+   */
+  public async putConsentId(
+    consentId: string,
+    requestBody: SDKStandardComponents.PutConsentsRequest,
+    destParticipantId: string
+  ): Promise<void> {
```
jgeewax, line 251: Also, why do we never want the results of these? Is there any? Do we care about those at all?

src/shared/ml-thirdparty-simulator/factories/consents.ts
```diff
@@ -0,0 +1,96 @@
...
+import * as faker from 'faker'
+
+import SDKStandardComponents from '@mojaloop/sdk-standard-components'
+
+export class ConsentFactory {
+  public static createPutConsentRequestIdRequest(
+    requestBody: SDKStandardComponents.PostConsentRequestsRequest
+  ): SDKStandardComponents.PutConsentRequestsRequest {
+    return {
+      authChannels: ['OTP'],
```
jgeewax, line 36: Isn't this Web by default?

src/shared/ml-thirdparty-client/index.ts
stevenwjy: Uhmm... this is only the internal implementation for the PISP demo server... Sorry if this looks a bit scary, but the reason is because the plugins to make request and the one used to simulate the callbacks from mojaloop are different (i.e., separated into `ml-thirdparty-client` and `ml-thirdparty-simulator` plugins). 

If you want to avoid having a variable like this, maybe it would be better to merge those plugins into one single plugin and make the variable private

src/shared/ml-thirdparty-simulator/factories/consents.ts
akapur99: Changed

src/shared/ml-thirdparty-client/index.ts
akapur99: Good point - I'm not sure actually - was just following what the rest of the repo was doing 
same with the awaiting 
so - now shifted to returning everything without await-ing and leaving it up to which ever method called this request to determine what to do with the promise
thanks!

src/shared/ml-thirdparty-simulator/factories/consents.ts
eoln, line 36: what about U2F ?

akapur99: at the moment I think we're sticking to Web and OTP? @stevenwjy

src/shared/ml-thirdparty-client/index.ts
stevenwjy: I think we still need to perform the await here to get back the result of the request (i.e., the response code). 
However, in the current implementation, as we still haven't handled the error cases, it's just awaiting for the response without doing anything. I personally feel that the function still need to run asynchronously and return a future instead of just invoking the async function and return directly.

src/shared/ml-thirdparty-simulator/factories/consents.ts
stevenwjy: Yeah, I believe `U2F` is not an option for the authentication by the DFSP based on the linking design [here](https://github.com/mojaloop/pisp/tree/master/docs/linking)

src/shared/ml-thirdparty-client/index.ts
akapur99: so essentially just separating the two classes completely? in which case we're just using simulator for testing and it wouldn't be part client library at all, right?

src/shared/ml-thirdparty-client/index.ts
```diff
@@ -102,59 +106,142 @@ export class Client {
...
+    return undefined
+  }
+
+  /**
+   * Gets a list of PISP/DFSP participants
+   */
+  public async getParticipants(): Promise<
+  SDKStandardComponents.GenericRequestResponse | undefined
+  > {
+    // TODO: Add once implemented in sdk-standard components
```
akapur99, line 161: throw NotImplementedYetError

src/shared/ml-thirdparty-client/index.ts
lewisdaly: I think the best approach would be to make a `Client` interface, and 2 classes which implement it `SimulatedClient` and `LiveClient` (or a better name...)

That way, you can pass through either a `SimulatedClient` which does the simulated calls or the real client which actually connects to Mojaloop

jgeewax (approved):

PR #23, opened 2020-08-27
Add script to perform URL registration

stevenwjy:


scripts/url-registration.ts
```diff
@@ -0,0 +1,182 @@
...
+  targetServiceId: string
+  type: string
+  endpointPath: string
+}
+
+const registrationValues: Array<RegistrationValue> = [
+  // PUT /participants
+  {
+    targetServiceId: 'central-ledger',
+    type: 'FSIOP_CALLBACK_URL_PARTICIPANT_PUT',
```
lewisdaly, line 39: I think we should use the `@mojaloop/central-services-shared` enums here

scripts/url-registration.ts
```diff
@@ -0,0 +1,182 @@
...
+  },
+  // PUT /transfers/{ID}/error
+  {
+    targetServiceId: 'central-ledger',
+    type: 'FSPIOP_CALLBACK_URL_TRANSFER_PUT_ERROR',
+    endpointPath: '/transfers/{{transferId}}/error',
+  }
+]
+
+program.version('0.0.1')
```
lewisdaly, line 104: Can this read from `package.json`?

scripts/url-registration.ts
```diff
@@ -0,0 +1,182 @@
...
+    return group.toUpperCase();
+  })
+}
+
+registrationValues.forEach(async (value) => {
+  // Compute variable name after being altered by the commander library
+  const variableName = toCamelCase('ml-' + value.targetServiceId)
+
+  // Target URL in Mojaloop
+  const targetUrl = program[variableName] + '/participants/'
```
lewisdaly, line 162: It's nicer to use `${program[variableName]}/participants/${program.participantId}/endpoints` string formatting

lewisdaly (changes requested):

scripts/url-registration.ts
stevenwjy: Ohh okay, will update this later

stevenwjy: Uhmm... sorry I think I might have some misunderstanding here... do you mean making the request to `central-services-shared`? Or use the enum defined there? Because I think the current enum definition is already based on the one defined in `@mojaloop/central-services-shared`

lewisdaly: I mean instead of hardcoding the strings, import the enums from central-services-shared and use those

stevenwjy: Ohhh I see... okay, got it! will update later 👍

stevenwjy: done :-)

scripts/url-registration.ts
```diff
@@ -0,0 +1,185 @@
...
+  {
+    targetServiceId: 'central-ledger',
+    type: EndPoints.FspEndpointTypes.FSPIOP_CALLBACK_URL_PARTICIPANT_PUT,
+    endpointPath: '/participants',
+  },
+  // PUT /parties/{Type}/{ID}
+  {
+    targetServiceId: 'central-ledger',
+    type: EndPoints.FspEndpointTypes.FSPIOP_CALLBACK_URL_PARTIES_PUT,
+    endpointPath: '/parties/{{partyIdType}}/{{partyIdentifier}}',
```
stevenwjy, line 50: I think the `FspEndpointTemplates` for this endpoint is missing @lewisdaly :o

scripts/url-registration.ts
```diff
@@ -0,0 +1,185 @@
...
+  // PUT /consents/{ID}/error
+  {
+    targetServiceId: 'central-ledger',
+    type: EndPoints.FspEndpointTypes.TP_CB_URL_CONSENT_PUT_ERROR,
+    endpointPath: EndPoints.FspEndpointTemplates.TP_CONSENT_PUT_ERROR,
+  },
+  // POST /authorizations
+  {
+    targetServiceId: 'central-ledger',
+    type: 'FSPIOP_CALLBACK_URL_TRX_REQ_SERVICE',
```
stevenwjy, line 91: not sure about this enum, does it exist?

scripts/url-registration.ts
```diff
@@ -0,0 +1,185 @@
...
+  // PUT /transfers/{ID}
+  {
+    targetServiceId: 'central-ledger',
+    type: EndPoints.FspEndpointTypes.FSPIOP_CALLBACK_URL_TRANSFER_PUT,
+    endpointPath: '/transfers/{{transferId}}',
+  },
+  // PUT /transfers/{ID}/error
+  {
+    targetServiceId: 'central-ledger',
+    type: 'FSPIOP_CALLBACK_URL_TRANSFER_PUT_ERROR',
```
stevenwjy, line 103: also not sure about both transfers endpoint

scripts/url-registration.ts
```diff
@@ -0,0 +1,185 @@
...
+  type: string
+  endpointPath: string
+}
+
+const registrationValues: Array<RegistrationValue> = [
+  // PUT /participants
+  {
+    targetServiceId: 'central-ledger',
+    type: EndPoints.FspEndpointTypes.FSPIOP_CALLBACK_URL_PARTICIPANT_PUT,
+    endpointPath: '/participants',
```
stevenwjy, line 44: also shall I use the value of `ORACLE_PARTICIPANTS_BATCH` to replace this @lewisdaly ?

scripts/url-registration.ts
eoln: yes /authorizations resised temporary in transaction-request-service

eoln (approved):
lgtm

PR #24, opened 2020-08-31
Handlers and Endpoints for Mojaloop/App/Firebase - Linking/Unlinking Flow

akapur99:
Addresses: https://app.zenhub.com/workspaces/mojaloop-project-59edee71d1407922110cf083/issues/mojaloop/project/1654

- [x] Implement all endpoint handlers + related domain logic, that deal with Mojaloop
- [x] Implement all handlers + related domain logic, that deal with Firebase (the app-side) 
- [x] Implement consent repository
- [x] Unit Tests 

Unit Tests for firestore repositories will be handles in a separate PR


src/models/consent.ts
```diff
@@ -40,18 +40,28 @@ export enum ConsentStatus {
    * Waiting for the user to confirm payee information and provide more
    * details about the transaction.
    */
-  PENDING_PAYEE_CONFIRMATION = 'PENDING_PAYEE_CONFIRMATION',
+  PENDING_PARTY_CONFIRMATION = 'PENDING_PARTY_CONFIRMATION',
```
eoln, line 43: I would keep PAYEE if it is more related to usage, PARTY is much more generic

src/models/consent.ts
```diff
@@ -40,18 +40,28 @@ export enum ConsentStatus {
...
 
   /**
    * Waiting for the user to authorize the consent.
    */
-  AUTHORIZATION_REQUIRED = 'AUTHORIZATION_REQUIRED',
+  AUTHENTICATION_REQUIRED = 'AUTHENTICATION_REQUIRED',
```
eoln, line 48: there is a difference between Autentication and Authorization, comment is telling about AUTHORIZATION of the consent so I would keep it as it is

src/server/handlers/firestore/consents.ts
```diff
@@ -0,0 +1,203 @@
...
+  // Check whether the consent document has all the necessary properties
+  // to perform a party lookup.
+  if (!validator.isValidPartyLookup(consent)) {
+    throw new Error('Consent Object Missing Fields')
+  }
+
+  // Payee is guaranteed to be non-null by the validator.
+
+  server.app.mojaloopClient.getParties(
+    // @ts-ignore
```
eoln, line 59: @ts-ignore should be banned - isn't it better to use casting instead so it will be much more clear what is going on here?

akapur99: used non-null insertion instead

src/server/handlers/openapi/mojaloop/parties/{Type}/{ID}.ts
```diff
@@ -30,37 +31,63 @@ import { PartiesTypeIDPutRequest } from '~/shared/ml-thirdparty-client/models/op
...
-      "status": Status.PENDING_PARTY_LOOKUP,
-    },
-    // Update the given field by their new values
-    {
-      payee: body.party,
-      status: Status.PENDING_PAYEE_CONFIRMATION,
-    }
-  )
+
+  if (partyIdType === 'OPAQUE') {
```
jgeewax, line 59: Add a note why this is OK. (Opaque is during linking when you're fetching the accounts available for linking from a pre-determined DFSP).

src/models/consent.ts
stevenwjy: I think this should be fine since this change is based on the sequence diagrams [here](https://github.com/mojaloop/pisp-demo-server/blob/master/docs/design/linking.md). Also, there's no payee involved when establishing a consent, so I suspect that the current code in the master is problematic... probably due to copying the enum used for the transfer flow before

src/models/consent.ts
stevenwjy: Same with my other comment, I think it's not an authorization since this is performed when establishing consent, not for authorizing a transfer. In this phase, the DFSP will ask user to perform authentication on their website or perhaps via OTP. Hence, I think the word "Authentication" better suits this case rather than "Authorization".

src/models/consent.ts
```diff
@@ -40,22 +40,37 @@ export enum ConsentStatus {
...
 
+  /**
+   * The consent is ACTIVE and challenge has been generated
+   */
+  CHALLENGE_GENERATED = 'CHALLENGE_GENERATED',
+
+  /**
+   * The consent is ACTIVE and challenge has been verified
+   */
+  CHALLENGE_VERIFIED = 'CHALLENGE_VERIFIED',
```
stevenwjy, line 63: Where is this used? Could you perhaps help to update the sequence diagrams if there are any changes for the design ...

src/repositories/participants.ts
```diff
@@ -0,0 +1,80 @@
...
+          // If previous participants list exists, delete it
+
+          // Iterate through all matching documents add them to the processing batch.
+          response.docs.forEach((doc) => {
+            batch.delete(doc.ref)
+          })
+        }
+        // Iterate through received participants list and add them to the processing batch.
+        data.array.forEach((participant: any) => {
+          batch.set(collectionRef.doc(), participant)
```
stevenwjy, line 68: I think it would be better to use `batch.add` rather than `batch.set` here

src/repositories/participants.ts
```diff
@@ -0,0 +1,80 @@
...
+   * Replace existing participants list with new list.
+   *
+   * @param data   Documents that are about to be added.
+   */
+  replace(data: Record<string, any>): Promise<void>
+}
+
+export class FirebaseParticipantRepository implements IParticipantRepository {
+  // TODO: Confirm data Type
+  async replace(data: Record<string, any>): Promise<void> {
```
stevenwjy, line 43: I think the parameter should be a list of participants instead of an arbitrary record

src/server/handlers/firestore/consents.ts
```diff
@@ -0,0 +1,228 @@
...
+    throw new Error('Consent Object Missing Fields')
+  }
+
+  // Party is guaranteed to be non-null by the validator.
+
+  server.app.mojaloopClient.getParties(
+    consent.party!.partyIdInfo.partyIdType,
+    consent.party!.partyIdInfo.partyIdentifier
+  )
+  //   }
```
stevenwjy, line 68: Perhaps can remove this line?

src/server/handlers/firestore/consents.ts
```diff
@@ -0,0 +1,228 @@
...
+
+    case ConsentStatus.AUTHENTICATION_REQUIRED:
+      await handleAuthentication(server, consent)
+      break
+
+    case ConsentStatus.ACTIVE:
+      await handleChallengeGeneration(server, consent)
+      break
+
+    case ConsentStatus.CHALLENGE_VERIFIED:
```
stevenwjy, line 220: When do you use this? Sorry I couldn't recall a phase that use this status ...

src/server/handlers/firestore/consents.validator.ts
```diff
@@ -0,0 +1,136 @@
...
+/**
+ * Checks whether a consent document has all the necessary fields to perform
+ * a party lookup.
+ *
+ * @param consent the object representation of a consent that is stored
+ *                    on Firebase.
+ */
+export const isValidPartyLookup = (consent: Consent): boolean => {
+  if (
+    consent?.party?.partyIdInfo &&
```
stevenwjy, line 37: Would it be better to use this format rather than the one used in `transactions.validator.ts`? Perhaps we better make both of them consistent... If you feel that this is much cleaner perhaps can also update the one in `transactions.validator.ts`?

src/server/handlers/openapi/mojaloop/consentRequests/{ID}.ts
```diff
@@ -20,14 +20,29 @@
...
+import { consentRepository } from '~/repositories/consent'
+import { ConsentStatus } from '~/models/consent'
 
-export const put: Handler = async (context: Context, request: Request, h: ResponseToolkit) => {
+export const put: Handler = async (
+  context: Context,
+  request: Request,
+  h: ResponseToolkit
+) => {
   logger.logRequest(context, request, h)
```
stevenwjy, line 38: I think we can remove the logging from master for this and other handlers

src/server/handlers/openapi/mojaloop/consentRequests/{ID}.ts
```diff
@@ -20,14 +20,29 @@
...
+
+  const id = context.request.params.id as string
```
stevenwjy, line 40: shouldn't it be `params.ID` instead of `params.id` (based on the open api definition)? Or perhaps have you checked whether this one works? iirc the variables are case sensitive ...

src/server/handlers/openapi/mojaloop/parties/{Type}/{ID}.ts
stevenwjy: +1, I think it would be better to add some comments in the code that explains about the use of opaque identifiers and how it is special

src/shared/ml-thirdparty-client/index.ts
```diff
@@ -209,10 +209,12 @@ export class Client {
    * Performs a request to generate a challenge for FIDO registration
    *
    * @param _consentId     identifier of consent as defined by Mojaloop API.
+   * @param destParticipantId   ID of destination - to be used when sending request
    */
   public async postGenerateChallengeForConsent(
     // eslint-disable-next-line @typescript-eslint/no-unused-vars
-    _consentId: string
+    _consentId: string,
```
stevenwjy, line 216: what's with the `_` at the beginning of each variable name? as far as I understand, parameter variables shouldn't be named like that right as it is usually used for private class variables?

test/unit/server/handlers/openapi/mojaloop/consentRequests/{ID}.test.ts
```diff
@@ -0,0 +1,125 @@
...
+} as unknown) as ResponseToolkit
+
+const postConsentRequestRequest: SDKStandardComponents.PostConsentRequestsRequest = {
+  id: '111',
+  initiatorId: 'pispA',
+  authChannels: ['WEB', 'OTP'],
+  scopes: [
+    {
+      accountId: 'as2342',
+      actions: ['account.getAccess', 'account.transferMoney'],
```
stevenwjy, line 70: is this new set of actions? could you help to update the sequence diagrams? thank you!

stevenwjy (changes requested):

src/repositories/participants.ts
akapur99: batch.add doesn't exist unfortunately

src/shared/ml-thirdparty-client/index.ts
akapur99: that's just because they're not being used currently - once implemented we need to remove that

src/server/handlers/openapi/mojaloop/consentRequests/{ID}.ts
akapur99: ah i was under the assumption they weren't sensitive, thanks for the catch

src/server/handlers/openapi/mojaloop/consentRequests/{ID}.ts
akapur99: makes sense

src/repositories/participants.ts
stevenwjy: Ohhh I see, didn't know about that... okay then :-)

src/shared/ml-thirdparty-client/index.ts
stevenwjy: Ohh I see, okay then :-)

src/server/handlers/openapi/mojaloop/consentRequests/{ID}.ts
akapur99: actually why not leave it in?

akapur99: wouldn't it make sense for a demo server to keep a record of transactions, linking etc.

test/unit/server/handlers/openapi/mojaloop/consentRequests/{ID}.test.ts
akapur99: yup - once I'm done with this PR - will open a new one to update diagrams throughout

src/interface/mojaloop.yaml
```diff
@@ -270,12 +270,17 @@ paths:
...
-      description: Result of deleting the specified consent
-      operationId: deleteConsentsById
+      summary: '[Mojaloop Callback] Result of patching the specified consent (used in unlinking)'
+      description: Result of patching the specified consent (used in unlinking)
+      operationId: patchConsentsById
+      requestBody:
+        content:
+          application/json:
+            schema:
+              $ref: 'shared.yaml#/components/schemas/PatchConsentsByIdRequest'
```
eoln, line 283: isn't it better to take this interface from api-snippets?

src/repositories/consent.ts
```diff
@@ -58,6 +79,53 @@ export class FirebaseConsentRepository implements IConsentRepository {
...
+      .firestore()
+      .collection('consents')
+
+    // Chain all of the given conditions to the query
+    for (const key in conditions) {
+      firestoreQuery = firestoreQuery.where(key, '==', conditions[key])
+    }
+
+    // Find and update all matching documents in Firebase that match the given conditions.
+    await firestoreQuery
```
eoln, line 104: it isn't good to mix `await` with `then` style for handling asynchronous operations: there is a risk of misunderstanding of what is going on here. Is it really needed?

src/repositories/participants.ts
```diff
@@ -0,0 +1,82 @@
...
+ * Mojaloop Foundation
+ - Name Surname <name.surname@mojaloop.io>
+
+ * Google
+ - Steven Wijaya <stevenwjy@google.com>
+ - Abhimanyu Kapur <abhi.kapur09@gmail.com>
+ --------------
+ ******/
+
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
eoln, line 27: drop this lint exception please, or you will be doomed ;-P

src/repositories/participants.ts
```diff
@@ -0,0 +1,82 @@
...
+}
+
+export class FirebaseParticipantRepository implements IParticipantRepository {
+  async replace(data: Participant[]): Promise<void> {
+    const collectionRef: FirebaseFirestore.CollectionReference<FirebaseFirestore.DocumentData> = firebase
+      .firestore()
+      .collection('participants')
+
+    // Find and update all matching documents in Firebase that match the given conditions.
+    await collectionRef
```
eoln, line 51: mixing `await` with `then` is antipattern

src/repositories/transaction.ts
```diff
@@ -24,6 +24,8 @@
  ******/
 
 /* eslint-disable @typescript-eslint/no-explicit-any */
```
eoln, line 26: drop this lint extension, please

src/server/handlers/firestore/consents.ts
```diff
@@ -0,0 +1,225 @@
...
+    consentRequestId: uuid.v4(),
+    status: ConsentStatus.PENDING_PARTY_LOOKUP,
+  })
+}
+
+async function handlePartyLookup(server: Server, consent: Consent) {
+  // Check whether the consent document has all the necessary properties
+  // to perform a party lookup.
+  if (!validator.isValidPartyLookup(consent)) {
+    throw new Error('Consent Object Missing Fields')
```
eoln, line 60: let use a specialized Error class instead generic Error with message

src/server/handlers/firestore/consents.ts
```diff
@@ -0,0 +1,225 @@
...
+}
+
+async function handleChallengeGeneration(server: Server, consent: Consent) {
+  if (!validator.isValidChallengeGeneration(consent)) {
+    throw new Error('Consent Object Missing Fields')
+  }
+
+  try {
+    server.app.mojaloopClient.postGenerateChallengeForConsent(
+      consent.consentId as string,
```
eoln, line 128: is this casting necessary?

src/server/handlers/firestore/consents.ts
```diff
@@ -0,0 +1,225 @@
...
+  if (!validator.isValidSignedChallenge(consent)) {
+    throw new Error('Consent Object Missing Fields')
+  }
+
+  try {
+    server.app.mojaloopClient.putConsentId(
+      consent.consentId as string,
+      {
+        requestId: consent.id,
+        initiatorId: consent.initiatorId as string,
```
eoln, line 146: are these casting necessary?

src/server/handlers/firestore/consents.ts
```diff
@@ -0,0 +1,225 @@
...
+      break
+
+    case ConsentStatus.ACTIVE:
+      await handleSignedChallenge(server, consent)
+      break
+
+    case ConsentStatus.REVOKE_REQUESTED:
+      await handleRevokingConsent(server, consent)
+      break
+  }
```
eoln, line 224: lack of the default case would bring problems in the future, let define default and throw an exception if reached

src/server/handlers/firestore/consents.validator.ts
```diff
@@ -0,0 +1,137 @@
...
+/**
+ * Checks whether a consent document has all the necessary fields to be
+ * processed as an authenticated consent request.
+ *
+ * @param consent the object representation of a consent that is stored
+ *                    on Firebase.
+ */
+export const isValidAuthentication = (consent: Consent): boolean => {
+  if (
+    consent.consentRequestId &&
```
eoln, line 55: same as above given by @stevenwjy

src/server/handlers/firestore/consents.validator.ts
```diff
@@ -0,0 +1,137 @@
...
+
+/**
+ * Checks whether a consent document has all the necessary fields to be
+ * processed as a request to generate challenge for a consent.
+ *
+ * @param consent the object representation of a consent that is stored
+ *                    on Firebase.
+ */
+export const isValidChallengeGeneration = (consent: Consent): boolean => {
+  if (consent.consentId && consent.party && consent.party.partyIdInfo?.fspId) {
```
eoln, line 98:
```suggestion
  return !!(consent?.consentId && consent.party?.partyIdInfo?.fspId)
```

eoln: maybe returning expression will fit better in place of if/else -> true/false mapping?

src/server/handlers/openapi/mojaloop/participants.ts
```diff
@@ -20,14 +21,25 @@
...
+import { participantRepository } from '~/repositories/participants'
 
-export const put: Handler = async (context: Context, request: Request, h: ResponseToolkit) => {
+export const put: Handler = async (
+  context: Context,
+  request: Request,
+  h: ResponseToolkit
+) => {
   logger.logRequest(context, request, h)
+  // @ts-ignore
```
eoln, line 39: don't use //ts-ignore use `as` to inform what type of value you are expecting here

src/server/handlers/openapi/mojaloop/consents/{ID}.ts
```diff
@@ -20,19 +20,39 @@
...
-export const put: Handler = async (context: Context, request: Request, h: ResponseToolkit) => {
+export const put: Handler = async (
+  context: Context,
+  request: Request,
+  h: ResponseToolkit
+) => {
   logger.logRequest(context, request, h)
+  // Updates consent fields
+  consentRepository.updateConsentById(
+    context.request.params.ID as string,
```
eoln, line 40: there is inconsistency of handling context.request.params.ID, sometimes there is inline casting, whereas in other cases const is declared (with casting) and used as function argument. Let keep consistency across all files

src/server/plugins/internal/firestore.ts
```diff
@@ -40,6 +47,11 @@ export interface Options {
       onUpdate?: TransactionHandler
       onRemove?: TransactionHandler
     }
+    consents: {
+      onCreate?: ConsentHandler
```
eoln, line 51: all optional properties smell bad -> later there is a need to always check is it present before a use

src/server/plugins/internal/firestore.ts
```diff
@@ -63,13 +78,59 @@ const listenToTransactions = (server: Server, options: Options): (() => void) =>
...
-
+          transactionHandlers.onUpdate(server, {
+            id: change.doc.id,
+            ...change.doc.data(),
+          })
         } else if (change.type === 'removed' && transactionHandlers.onRemove) {
-          transactionHandlers.onRemove(server, { id: change.doc.id, ...change.doc.data() })
+          transactionHandlers.onRemove(server, {
+            id: change.doc.id,
+            ...change.doc.data(),
```
eoln, line 93: copy all properties could be dangerous in schema-free databases - there is a possibility to store data we don't want/need to store

src/shared/ml-thirdparty-simulator/factories/transfer.ts
```diff
@@ -28,21 +28,24 @@ import * as faker from 'faker'
...
+   *
    * @param _               transaction request id of the corresponsing authorization.
    * @param __              an authorization object as defined by the Mojaloop API.
    * @param transactionId   transaction id to be associated with the transfer object.
    */
-  public static createTransferIdPutRequest(_: string,
-    __: AuthorizationsPutIdRequest, transactionId: string): TransferIDPutRequest {
+  public static createTransferIdPutRequest(
+    _: string,
+    __: AuthorizationsPutIdRequest,
```
eoln, line 46: lovely magic of underscores.... do you know that the floors in debugging hell are full of underscores?

test/unit/server/handlers/openapi/mojaloop/consentRequests/{ID}.test.ts
```diff
@@ -0,0 +1,125 @@
...
+  },
+}))
+
+// Mock firebase to prevent transaction repository from opening the connection.
+jest.mock('~/lib/firebase')
+
+const mockRequest = jest.fn().mockImplementation()
+
+const mockResponseToolkit = ({
+  response: (): ResponseObject => {
```
eoln, line 54: again bad, very bad mocking of responseToolkit (please check you copy-paste clipboard and clean it up)

test/unit/server/handlers/openapi/mojaloop/consents.test.ts
```diff
@@ -0,0 +1,130 @@
...
+    logRequest: jest.fn().mockImplementation(),
+  },
+}))
+
+// Mock firebase to prevent transaction repository from opening the connection.
+jest.mock('~/lib/firebase')
+
+const mockRequest = jest.fn().mockImplementation()
+
+const mockResponseToolkit = ({
```
eoln, line 53: oh Gosh, did you make it again?

eoln (changes requested):
- don't use lint-exception for `any`
- use specialized Error classes when throwing exceptions
- refactor if/else -> true/false mapping using expressions with proper null-coalescing operator: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html
- try to not use `as` casting operator so often -> check if interfaces are declared correctly
- there is a need to improve the style to not use so many underscores

src/server/plugins/internal/firestore.ts
akapur99: hi pawel - can you raise an issue for this - i just made style changes here and for the sake of separation of concerns I don't want to modify code not related to this ticket/PR

src/shared/ml-thirdparty-simulator/factories/transfer.ts
akapur99: hi pawel - can you raise an issue for this - i just made style changes here and for the sake of separation of concerns I don't want to modify code not related to this ticket/PR

src/server/plugins/internal/firestore.ts
akapur99: hi pawel - can you raise an issue for this - i just made style changes here and for the sake of separation of concerns I don't want to modify code not related to this ticket/PR

src/shared/ml-thirdparty-simulator/factories/transfer.ts
eoln: https://github.com/mojaloop/pisp-demo-server/issues/26

src/repositories/consent.ts
akapur99: doesn't seem like it
@stevenwjy - what do you think?

src/lib/config.ts
```diff
@@ -108,6 +108,13 @@ const config = convict({
       default: 'pisp',
       env: 'MOJALOOP_PARTICIPANT_ID',
     },
+    // TODO: Replace placeholder
+    callbackUri: {
```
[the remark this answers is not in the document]
akapur99: @jgeewax  @kkzeng  @stevenwjy  - would a good name be `appCallbackUri` or `pispCallbackUri` ?

src/server/plugins/internal/firestore.ts
akapur99: @stevenwjy is there a reason we made the oncreate and onupdate optional?

src/interface/mojaloop.yaml
stevenwjy: Yup, I think it would be better that way! :-) 
iirc, the current dependency still points to my forked repo (because it was set before the `api-snippets` was published), probably can fix that as well

src/lib/config.ts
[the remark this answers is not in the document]
stevenwjy: Hmm... actually I'm not really sure about this... just to confirm the callback here is the one used by the DFSP once it has successfully authenticated the user for account linking right?

src/repositories/consent.ts
stevenwjy: oh yeah, sorry didn't notice about this... I think we can just remove the `await`? Does this occur in the code for the transfer repository as well? maybe can consider removing that ambiguity as well

src/server/handlers/firestore/consents.validator.ts
stevenwjy: Oh right I think this is a better way to do it rather than using if/else or `!= null` :o

src/server/handlers/openapi/mojaloop/consentRequests/{ID}.ts
stevenwjy: not sure about this, but personally I don't think its a good idea to just log everything in a function like this (which will be called so many times)... I usually prefer to just record some critical parts or if there is any error... how about other mojaloop services? do they log every requests?

src/interface/mojaloop.yaml
akapur99: make sense - I'll make a separate ticket for that - separation of concerns

declarations.d.ts
```diff
@@ -0,0 +1,27 @@
+/*****
```
eoln, line 1: no need for a separated declarations file, if you want to use central-services-shared, copy-paste declarations from thirdparty-[api/scheme]-adapter if you plan to use something which is missed there you need to declare it properly. Empty declarations only postpone in time the moment when declarations have to be written. Empty declarations allow to import module in TS but you can't use imported elements without TS complaints or complex casting.

eoln (changes requested):

akapur99:
> * don't use lint-exception for `any` - DONE
> * use specialized Error classes when throwing exceptions - DONE
> * refactor if/else -> true/false mapping using expressions with proper null-coalescing operator: - Done in one place but a tad confused where all I should be doing this? 
 https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html
> * try to not use `as` casting operator so often -> check if interfaces are declared correctly - Where exactly?
> * there is a need to improve the style to not use so many underscores - ummm

src/server/plugins/internal/firestore.ts
stevenwjy: Hmm, actually there's no specific reason... Since it is a plugin, I just thought to make it as customizable as possible for people who use the plugin, which means they could decide to not handle the create operation etc. This is like the implementation of some button widgets in the `pisp-demo-server-app`, typically the `onTap` parameter is optional, in case the button shouldn't do anything when people tap it (probably @kkzeng can clarify). However, considering that this is currently an internal plugin, I shouldn't be an issue to change it into a required option.

src/server/handlers/firestore/consents.validator.ts
```diff
@@ -0,0 +1,137 @@
...
+ * processed as an authenticated consent request.
+ *
+ * @param consent the object representation of a consent that is stored
+ *                    on Firebase.
+ */
+export const isValidAuthentication = (consent: Consent): boolean => {
+  if (
+    consent.consentRequestId &&
+    consent.consentId &&
+    consent.party &&
```
akapur99, line 57: switch to optional chaining across

src/server/plugins/internal/firestore.ts
akapur99: @jgeewax  any thoughts?

src/lib/config.ts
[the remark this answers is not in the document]
akapur99: @jgeewax any specific preference?

src/server/plugins/internal/firestore.ts
jgeewax: I'm not 100% sure about this. Optional isn't necessarily bad, particularly when we actively just want "the default behavior". In this case, this same library is the only one accessing this property. And, I'm no TS expert, but I thought as of TS 3.7 you had the safe navigation/calling operator?

I'm talking about the `?.` operator:
  - `foo?.bar` will be `null` if `foo` is `null`
  - `foo.bar?.()` will call `foo.bar()` if it's not null, and do nothing if it's `null`

```typescript
if (change.type === 'added') {
  consentHandlers.onCreate?.(server, {
    id: change.doc.id,
    ...change.doc.data(),
  })
}
```

jgeewax (commented):
Just a couple of comment replies for now.

akapur99:
Wrt https://github.com/mojaloop/pisp-demo-server/pull/24#discussion_r487826431
@jgeewax  @eoln  I think using optional chaining makes sense and leaving it as optional. When i implemented that, some errors came up, which then also made me realize there's no tests for the firestore file at all. Hence I've opened a new issue for this and will take that on post this PR

src/models/errors.ts
```diff
@@ -37,3 +37,8 @@ export class MissingConsentFieldsError extends Error {
     this.consent = consent
   }
 }
+export class InvalidConsentStatusError extends Error {
+  public constructor(consentStatus: string) {
```
eoln, line 41: I suggest adding also as a parameter the consent.id so you can attach it to the error message which will be helpful later with debugging

src/repositories/consent.ts
```diff
@@ -19,27 +19,47 @@
...
 import { logger } from '~/shared/logger'
 
 export interface IConsentRepository {
+  /**
+   * Updates a consent document based on a unique identifier.
+   *
+   * @param id    Id for the consent document that needs to be updated.
+   * @param data  Document fields that are about to be updated.
+   */
+  updateConsentById(id: string, data: Record<string, unknown>): Promise<void>
```
eoln, line 40: shouldn't we use the Consent interface as data type here? Record<string, unknown> is very generic

src/repositories/consent.ts
```diff
@@ -19,27 +19,47 @@
...
+
+  /**
+   * Updates one or more consent documents based on the given conditions.
+   *
+   * @param conditions  Conditions for the documents that need to be updated.
+   * @param data        Document fields that are about to be updated.
+   */
+  updateConsent(
+    conditions: Record<string, unknown>,
+    data: Record<string, unknown>
```
eoln, line 57: let use Consent interface here

src/repositories/consent.ts
```diff
@@ -58,6 +78,53 @@ export class FirebaseConsentRepository implements IConsentRepository {
...
+          batch.update(
+            // Put a reference to the document.
+            firebase.firestore().collection('consents').doc(doc.id),
+            // Specify the updated fields and their new values.
+            data
+          )
+        })
+
+        // Commit the updates.
+        return batch.commit()
```
eoln, line 122: if batch.commit() returns a Promise where it is resolved or where rejection is handled?

src/repositories/participants.ts
```diff
@@ -0,0 +1,81 @@
...
+            batch.delete(doc.ref)
+          })
+        }
+        // Iterate through received participants list and add them to the processing batch.
+        data.forEach((participant: Participant) => {
+          batch.set(collectionRef.doc(), participant)
+        })
+
+        // Commit the updates.
+        return batch.commit()
```
eoln, line 73: where this Promise is handled?

src/repositories/transaction.ts
```diff
@@ -35,7 +36,7 @@ export interface ITransactionRepository {
    * @param id    Id for the transaction document that needs to be updated.
    * @param data  Document fields that are about to be updated.
    */
-  updateById(id: string, data: Record<string, any>): Promise<void>
+  updateById(id: string, data: Record<string, unknown>): Promise<void>
```
eoln, line 39: don't we have a Transaction interface defined?

src/server/handlers/firestore/consents.ts
```diff
@@ -0,0 +1,225 @@
...
+    throw new MissingConsentFieldsError(consent)
+  }
+
+  // Fields are guaranteed to be non-null by the validator.
+  try {
+    server.app.mojaloopClient.getParties(
+      consent.party!.partyIdInfo.partyIdType,
+      consent.party!.partyIdInfo.partyIdentifier
+    )
+  } catch (error) {
```
eoln, line 65: I see this pattern everywhere -> the thrown error is logged only and exceptions aren't handled - is it intentionally? if Yes, please add some explanation in comment to hint reader.

I think the name of function `handle` is misleading - here I see that async mojaloop API is called and somewhere else we will wait on our PUT endpoint to receive requested information -> so maybe instead `handle` use `initiate` and add in comments where I can find the code which is handling the PUT response...

src/server/handlers/firestore/consents.validator.ts
```diff
@@ -0,0 +1,105 @@
...
+ * Checks whether a consent document has all the necessary fields to perform
+ * a party lookup.
+ *
+ * @param consent the object representation of a consent that is stored
+ *                    on Firebase.
+ */
+export const isValidPartyLookup = (consent: Consent): boolean => {
+  return !!(
+    consent?.party?.partyIdInfo?.partyIdType &&
+    consent.party.partyIdInfo.partyIdentifier
```
eoln, line 38: don't we need to use in the second part of the expression the chaining operator?

src/server/handlers/openapi/mojaloop/consentRequests/{ID}.ts
```diff
@@ -20,14 +20,25 @@
...
 
-export const put: Handler = async (context: Context, request: Request, h: ResponseToolkit) => {
-  logger.logRequest(context, request, h)
+export const put: Handler = async (
+  context: Context,
+  _request: Request,
+  h: ResponseToolkit
+) => {
+  const { authChannels, authUri } = context.request.body
+  consentRepository.updateConsentById(context.request.params.ID as string, {
```
eoln, line 38: let add a comment here we are not waiting for a promise to resolve so this part of the code is executed async

src/server/handlers/openapi/mojaloop/consents.ts
```diff
@@ -20,14 +20,27 @@
...
-export const post: Handler = async (context: Context, request: Request, h: ResponseToolkit) => {
-  logger.logRequest(context, request, h)
+export const post: Handler = async (
+  context: Context,
+  _request: Request,
+  h: ResponseToolkit
+) => {
+  const { id, initiatorId, participantId, scopes } = context.request.body
+
+  consentRepository.updateConsentById(id, {
```
eoln, line 39: let add not that we are not waiting on Promise to resolve...

src/server/handlers/openapi/mojaloop/consents/{ID}.ts
```diff
@@ -20,19 +20,36 @@
...
+import { consentRepository } from '~/repositories/consent'
 
-export const put: Handler = async (context: Context, request: Request, h: ResponseToolkit) => {
-  logger.logRequest(context, request, h)
+export const put: Handler = async (
+  context: Context,
+  _request: Request,
+  h: ResponseToolkit
+) => {
+  // Updates consent fields
```
eoln, line 36: let add comment we are not waiting on promise to resolve...

eoln (changes requested):

src/server/handlers/firestore/consents.validator.ts
akapur99: if the first part returned falsy at any point then the expression would short circuit before reaching the second part, no?

src/repositories/consent.ts
akapur99: Hi - this has been done intentionally as we are not passing in the entire consent object everytime - we pass in just the key/values that need to be updated

src/repositories/consent.ts
akapur99: same as above

src/repositories/transaction.ts
akapur99: same reason as consents

src/server/handlers/firestore/consents.ts
akapur99: makes sense 
@jgeewax  @stevenwjy  - was wondering if a specific protocol for error handling was discussed ? 

Also @eoln - I see your point re:  `handle` and have changed accordingly, but don't understand what you mean re: 'add in comments where I can find the code which is handling the PUT response...'

src/server/handlers/firestore/consents.validator.ts
eoln: yes you are right second part of AND will not be evaluated if the first part is already falsy :-D

src/repositories/consent.ts
eoln: if we pass fields that aren't present in the Consent interface - Will they be blindly stored? What happens if by mistake we will pass property of a different type instead?

src/repositories/consent.ts
akapur99: good question. Actually I'm not completely sure where it should be handled. The endpoint handlers which call this method typically call it async and then send back a response - so probably shouldn't be handling there 
we could simply add a try catch or .then.catch - with the .then() returning void and .catch throwing/logging error?

src/server/handlers/firestore/consents.ts
eoln: I think about the short hint, something like this ->` the corresponding put handler is implemented in file abc.ts`

akapur99: ahh makes sense

src/repositories/consent.ts
akapur99: Any external objects should be getting validated by the JOI validation + for firestore handlers - that's coming from the db itself - only other place is when we process internally and update
but still I agree we need to be more defensive here - maybe I can add in a check somewhere 
@stevenwjy was wondering if you have any thoughts

src/server/handlers/firestore/consents.ts
stevenwjy: I think that's a good idea -- it was previously just logged because we had not discussed much about proper error handling and hence the code tried to focus on the happy path

src/repositories/consent.ts
stevenwjy: The reason for not using the `Consent` interface was because the interface has some non-optional variables, which may not be suitable for performing the update since we only want to update some specific properties which may or may not be optional... But I agree that we should be more defensive here

src/server/handlers/firestore/consents.ts
akapur99: If we haven't maybe it would be a good time to make a ticket for it and reference that in comments everywhere we are logging an error that this is to be replaced?

src/repositories/consent.ts
```diff
@@ -102,25 +102,28 @@ export class FirebaseConsentRepository implements IConsentRepository {
...
+              data
+            )
+          })
 
-        // Commit the updates.
-        return batch.commit()
-      })
+          // Commit the updates.
+          await batch.commit()
+          return undefined
```
eoln, line 124: no need for `return`

src/repositories/transaction.ts
```diff
@@ -88,7 +88,8 @@ export class FirebaseTransactionRepository implements ITransactionRepository {
         })
 
         // Commit the updates.
-        return batch.commit()
+        await batch.commit()
+        return undefined
```
eoln, line 92: no need for return

eoln (changes requested):

src/server/handlers/firestore/consents.ts
akapur99: Following up - will do as stated above

src/repositories/consent.ts
akapur99: Writing an issue on this 
We should ensure firebase security rules are in place and do as much as we can with those, else log issues - and can run an internal validator on it

akapur99: https://github.com/mojaloop/pisp-demo-server/issues/new

src/server/handlers/firestore/consents.ts
akapur99: https://github.com/mojaloop/pisp-demo-server/issues/33

src/server/handlers/openapi/mojaloop/index.ts
```diff
@@ -41,7 +42,7 @@ export const apiHandlers = {
   postAuthorizations: MojaloopAuthorizations.post,
   postConsents: MojaloopConsents.post,
   putConsentsById: MojaloopConsentsById.put,
-  deleteConsentsById: MojaloopConsentsById.remove,
```
kkzeng, line 44: I'm guessing the patch is due to fact we don't want to remove consents but rather change their status to 'revoked'?

src/server/plugins/internal/firestore.ts
```diff
@@ -89,6 +100,48 @@ const listenToTransactions = (
...
+  options: Options
+): (() => void) => {
+  const consentHandlers = options.handlers.consents
+
+  return firebase
+    .firestore()
+    .collection('consents')
+    .onSnapshot((querySnapshot) => {
+      querySnapshot.docChanges().forEach((change) => {
+        if (change.type === 'added' && consentHandlers.onCreate) {
```
kkzeng, line 125: minor nit: are there no enums/constants to compare change type - why are comparing change.type to a hardcoded string

kkzeng (approved):
lgtm - just a nit and question. Would wait on those who requested changes to approve though before merging

src/server/plugins/internal/firestore.ts
```diff
@@ -89,6 +100,48 @@ const listenToTransactions = (
...
+          consentHandlers.onCreate(server, {
```
kkzeng, line 126: Should we not check whether or not these handlers are defined?

src/server/handlers/openapi/mojaloop/index.ts
akapur99: yup - but the app is free to do differently I believe

src/server/plugins/internal/firestore.ts
akapur99: I see your point - so I copied this portion from stevent's implementation of the `transactions` listener and I think steven copied this directly from the firebase docs
no harm in being more struct here with an enum, but given the limited scope - not sure its necessary

akapur99: leaving this here for @eoln 's thoughts

src/server/plugins/internal/firestore.ts
akapur99: we are - in the `if` statement
i'm not sure i understand what you mean

jgeewax (approved):

src/repositories/participants.ts
```diff
@@ -0,0 +1,79 @@
...
+}
+
+export class FirebaseParticipantRepository implements IParticipantRepository {
+  async replace(data: Participant[]): Promise<void> {
+    try {
+      const collectionRef: FirebaseFirestore.CollectionReference<FirebaseFirestore.DocumentData> = firebase
+        .firestore()
+        .collection('participants')
+
+      // Find and update all matching documents in Firebase that match the given conditions.
```
eoln, line 50: lack of condition

eoln (changes requested):

PR #28, opened 2020-09-10
Add mock data documentation for account linking related things

kkzeng:


README.md
```diff
@@ -79,3 +83,51 @@ Go to the PISP demo app and try to send money to a payee. You should see that th
...
+## Mocking Accounts for Associated Accounts
+
+1. In the Consent object, that you are working with, add an `accounts` field of type 'array'.
+2. In that array, add as many Account documents as you want.
+
+An Account document consists of a Map that looks like so:
+
+```
+{
+  id: "fspId",
```
eoln, line 116: `Account.id` suggests that `id` is of an `Account` but it is FSP.id -> I suggest to use `fspId` instead

eoln: or maybe in this example, `fspId` or `fspA` means Account's Id allocated/handled by given FSP -> so better is to use more descriptive values in this example...

kkzeng: That makes sense @eoln 
I will change the value.
The property names are kind of set already, I think I mistakenly wrote fspId, the id should actually be the id of the account.

kkzeng: I changed them to be similar to those in the linking diagrams. What do you think?

eoln (approved):

PR #30, opened 2020-09-11
Feat/Deciding + Implementing Ambient Declarations

akapur99:
Addresses: https://app.zenhub.com/workspaces/mojaloop-project-59edee71d1407922110cf083/issues/mojaloop/project/1696

### DIFF
--- a/ambient.d.ts
+++ b/ambient.d.ts
@@ -28,12 +28,12 @@ type Client = import('~/shared/ml-thirdparty-client').Client
 type Server = import('@hapi/hapi').Server
 type ServerApplicationState = import('@hapi/hapi').ServerApplicationState
 
-interface NewServerApplicationState extends ServerApplicationState {
+interface CustomServerApplicationState extends ServerApplicationState {
   mojaloopClient: Client
 }
 
 interface StateServer extends Server {
-  app: NewServerApplicationState
+  app: CustomServerApplicationState
 }
 
 // declare module '@mojaloop/sdk-standard-components' {
--- a/ambient.d.ts
+++ b/ambient.d.ts
@@ -35,11 +35,3 @@ interface PISPServerApplicationState extends ServerApplicationState {
 interface StateServer extends Server {
   app: PISPServerApplicationState
 }
-
-// declare module '@mojaloop/sdk-standard-components' {
-//   interface response {
-//     statusCode: number
-//     headers: Record<string, string>
-//     data: Record<string, unknown>
-//   }
-// }
--- a/ambient.d.ts
+++ b/ambient.d.ts
@@ -19,22 +19,27 @@
  - Name Surname <name.surname@mojaloop.io>
 
  * Google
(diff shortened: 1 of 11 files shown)


ambient.d.ts
```diff
@@ -19,22 +19,27 @@
...
-import { Client } from '~/shared/ml-thirdparty-client'
+type Client = import('~/shared/ml-thirdparty-client').Client
+type Server = import('@hapi/hapi').Server
+type ServerApplicationState = import('@hapi/hapi').ServerApplicationState
 
-declare module '@hapi/hapi' {
-  interface ServerApplicationState {
-    mojaloopClient: Client
-  }
+interface NewServerApplicationState extends ServerApplicationState {
```
stevenwjy, line 31: I think it would be better to name it `LocalServerApplicationState` or `CustomServerApplicationState`

eoln: `MojaServerApplicationState`

eoln: or `PISPServerApplicationSteate`

ambient.d.ts
```diff
@@ -19,22 +19,27 @@
...
-declare module '@mojaloop/sdk-standard-components' {
-  interface response {
-    statusCode: number
-    headers: Record<string, string>
-    data: Record<string, unknown>
-  }
+interface StateServer extends Server {
+  app: CustomServerApplicationState
 }
+
```
eoln, line 38: don't keep commented code, drop it

src/server/create.ts
```diff
@@ -41,9 +44,9 @@ export default async function create(config: ServiceConfig): Promise<Server> {
         failAction: onValidateFail,
       },
     },
-  })
+  }) as StateServer
```
eoln, line 47: no need for this casting

src/server/create.ts
```diff
@@ -41,9 +44,9 @@ export default async function create(config: ServiceConfig): Promise<Server> {
...
 
-  server = await plugins.register(server)
+  server = (await plugins.register(server)) as StateServer
```
eoln, line 49:
```suggestion
  await plugins.register(server)
```

src/server/create.ts
```diff
@@ -41,9 +44,9 @@ export default async function create(config: ServiceConfig): Promise<Server> {
...
   extensions.register(server)
 
   return server
```
eoln, line 52:
```suggestion
  return server as StateServer
```

eoln (changes requested):
small changes requested

ambient.d.ts
akapur99: I'm not totally sure why it exists - raman implemented it but doesn't seem to serve a purpose anymore
so i want to keep it around in case there's some future use?

ambient.d.ts
akapur99: that makes more sense yes

eoln (approved):
lgtm

PR #52, opened 2021-03-02
feat: reenable thirdparty simulator

lewisdaly:


package.json
```diff
@@ -115,7 +115,7 @@
     "faker": "^4.1.0",
     "firebase-admin": "^9.2.0",
     "openapi-backend": "^3.5.1",
-    "uuid": "^8.2.0",
+    "uuid": "^8.3.2",
```
eoln, line 118: I would move mojaloop/api-snippets dependency to devDependencies

src/server/handlers/firestore/transactions.ts
```diff
@@ -100,6 +100,7 @@ async function handlePartyConfirmation(
         transaction.consentId!
       )
 
+      //@ts-ignore - TODO Implement
```
eoln, line 103: `ts-ignore` is the Evil

src/shared/ml-thirdparty-client/hapi-plugin.ts
```diff
@@ -28,6 +28,11 @@
 import Client from '~/shared/ml-thirdparty-client'
 import { Plugin, Server } from '@hapi/hapi'
 import { Options } from './options'
+import config from '~/lib/config'
+// import { logger } from '~/shared/logger'
+import { Simulator } from '../ml-thirdparty-simulator'
```
eoln, line 33: having Simulator branch code in deployment isn't a good idea, we should get rich of it in the future....

src/shared/ml-thirdparty-simulator/hapi-plugin.ts
```diff
@@ -1,53 +1,54 @@
...
+//  * Google
+//  - Steven Wijaya <stevenwjy@google.com>
+//  - Abhimanyu Kapur <abhi.kapur09@gmail.com>
+//  --------------
+//  ******/
 
-import { Plugin, Server } from '@hapi/hapi'
-import { Simulator } from '~/shared/ml-thirdparty-simulator'
-import { Options } from './options'
+// import { Plugin, Server } from '@hapi/hapi'
```
eoln, line 27: why we need to keep the commented code - if not used let get rid of it

eoln (approved):
there is some commented code with Tech Debt. Let add some stories to the board to not forget about ;-)

PR #58, opened 2021-03-25
feat: use api-snippets(thirdparty-api paths) instead of local defs

sridharvoruganti:
related to #[2089](https://github.com/mojaloop/project/issues/2089) 

@lewisdaly @eoln 

Created draft PR for discussion, pisp-demo-server api definitions not in sync with api-snippets.

Existing API: https://github.com/mojaloop/pisp-demo-server/blob/master/src/interface/mojaloop.yaml
                     https://github.com/mojaloop/pisp-demo-server/blob/master/src/interface/shared.yaml

1.   /participants: endpoint is PUT without {ID}
request body contains participants[] { fspId:FspId, name:string }
is it equivalent to   /participants/{Type}/{ID}: participants_Type_ID
2. /parties/{Type}/{ID}:
 request body contains party (schemas/Party) and accounts(Account[] {id,currency})
where as PartiesTypeIDPutResponse defined in api-snippets contains only party
3. /accounts/{ID}:
body contains Account[] {id,currency}
where as api-snippets def contains []{accountNickname, id, currency}
4. /consents:
body contains extra fields initiatorId , participantId and credential , I understand initiatorId , participantId not required as per new design but what about credentials
5. /thirdpartyRequests/transactions/{ID}:
put: transactionId missing from request 
patch: body contains TransfersIDPutResponse?


lewisdaly:
Will be covered in mojaloop/project#2263

PR #62, opened 2021-06-04
feat: support web login flow demos

lewisdaly:


src/interface/shared.yaml
```diff
@@ -57,20 +57,31 @@ components:
...
         - VERIFIED
         - PENDING
+    CredentialPayload:
+      title: CredentialPayload
+      type: object
+      properties:
+        id: 
+          type: string
+        rawId:
+          type: array
```
eoln, line 67: array or base64 encoded string.... ?

src/models/demoAccount.ts
```diff
@@ -26,7 +26,7 @@ export interface DemoAccount {
   /**
    * The id of the credential associated with this account
    */
-  keyHandleId: Array<number>
+  keyHandleId: Uint8Array
```
eoln, line 29: this is optional field

lewisdaly:
This will be superseded by another PR soon

PR #63, opened 2021-06-15
feat(types): switch from custom types to api-snippets defined types

lewisdaly:
* switches from types defined in sdk to api-snippets
* updates the thirdparty api to use api-snippets
* replaces old API calls with updated calls (e.g. `PATCH /consentRequests/{id}`)

part of mojaloop/project#2263


src/interface/api_template/components/schemas/Participant.yaml
```diff
@@ -0,0 +1,8 @@
+title: Participant
```
kleyow, line 1: hmm. you shouldn't need to rewrite these schemas if they are apart of the fspiop/thirdparty and in snippets. is there some schema mismatch that I'm not seeing?

src/interface/api_interfaces/openapi.d.ts
```diff
@@ -0,0 +1,1698 @@
+/**
```
kleyow, line 1: you shouldn't need to regenerate data types if they are apart of fspiop/thirdparty.

lewisdaly: Ah ok, yeah I don't think we are referring to this file, so can probably just delete it, and remove it from the build task.

src/interface/api_template/components/schemas/Participant.yaml
lewisdaly: Yeah this was leftover from Sridhar's original work updating the types and moving from what our interns had already written. I suspect I should be able to delete a lot of these components and update them to references to api-snippets.

I'll take care of that in a followup PR (see the todos in `mojaloop-template.yaml`)

src/server/handlers/firestore/transactions.ts
```diff
@@ -157,7 +158,7 @@ async function handleAuthorization(server: StateServer, transaction: Transaction
 
   // TD - LD eww so much messy casting going on
   // @ts-ignore - todo fix me!
-  const requestBody: PutThirdpartyRequestsTransactionsAuthorizationsRequest = {
+  const requestBody: tpAPI.Schemas.ThirdpartyRequestsTransactionsIDAuthorizationsPutResponsequest = {
```
kleyow, line 161: typo?

lewisdaly: Haha yeah I guess the `ts-ignore` should be removed and we would see a built time error.

kleyow (approved):

PR #64, opened 2021-06-23
feat(api): continue cleaning up old API defs

lewisdaly:
- also fixes minor issues with account linking flows

part of mojaloop/project#2263


kleyow (approved):
lgtm

PR #65, opened 2021-06-24
feat(demo): tweaks to allow web and otp login flows

lewisdaly:


src/lib/config.ts
```diff
@@ -116,9 +116,22 @@ const config = convict({
     doc:
       'An ISO 4217 currency code to replace other values with for demo purposes',
     format: '*',
-    default: 'XTS',
+    default: 'TZS',
     env: 'DEMO_CURRENCY',
   },
+  simulatorDefaultAuthChannel: {
+    doc: 'The auth channel for the local simulator to use. Defaults to OTP',
```
kleyow, line 123: default doesn't match with description

kleyow (approved):

src/lib/config.ts
lewisdaly: Well spotted. I'll get that in the next one.

